PR #13, opened 2019-01-05
iss12_add adjust_slope_offset function

stephenholleran:
function to adjust a wind speed with a new slope and offset.

I have written lots of tests in a jupyter notebook. These will need to be properly implemented in the test area.

### DIFF
--- a/brightwind/transform/transform.py
+++ b/brightwind/transform/transform.py
@@ -19,7 +19,7 @@ import pandas as pd
 import math
 from ..utils import utils
 
-__all__ = ['average_data_by_period', 'scale_wind_speed', 'offset_wind_direction']
+__all__ = ['average_data_by_period', 'adjust_slope_offset', 'scale_wind_speed', 'offset_wind_direction']
 
 
 def _compute_wind_vector(wspd, wdir):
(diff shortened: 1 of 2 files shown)


Inder95:
Hi Stephen, I have added a basic skeleton to add tests for this function and pushed it to the repo. Do check it out. Also I followed this link to set up testing environment on my laptop:
http://blog.thedigitalcatonline.com/blog/2015/05/13/python-oop-tdd-example-part1/

brightwind/transform/transform.py
```diff
@@ -156,7 +156,46 @@ def average_data_by_period(data: pd.Series, period, aggregation_method='mean', f
...
+        df = bw.load_campbell_scientific(bw.datasets.demo_site_data)
+        df['Spd80mS_adj'] = bw.adjust_slope_offset(df.Spd80mS, 0.044, 0.235, 0.04365, 0.236)
+        df[['Spd80mS', 'Spd80mS_adj']]
+
+    """
+    try:
+        return new_slope * ((wspd - current_offset) / current_slope) + new_offset
+    except TypeError as type_error:
+        for arg_value in [current_slope, current_offset, new_slope, new_offset]:
+            if type(arg_value) is not float:
```
Inder95: They can be of type int too.

brightwind/transform/transform.py
```diff
@@ -156,7 +156,46 @@ def average_data_by_period(data: pd.Series, period, aggregation_method='mean', f
...
+                raise TypeError("argument '" + str(arg_value) + "' is not of data type number")
+        if type(wspd) is not int or type(wspd) is not float:
+            if type(wspd) == pd.DataFrame and (wspd.dtypes == object)[0]:
```
Inder95: wspd can be data series too.

brightwind/transform/transform.py
stephenholleran: added this

brightwind/transform/transform.py
stephenholleran: added Series option too

stephenholleran:
I've added those two items about int and Series.

Once you have accepted this pull request i'll add in the tests as i'd say I am going to get conflicts if I modify your test_transform.py file. Once I get familiar with this process I'll do it.

PR #33, opened 2019-03-05
Iss20 time continuity gaps docs

Inder95:
@stephenholleran  Please review the docstring and let me know if there are any suggestions.


brightwind/analyse/analyse.py
```diff
@@ -326,17 +326,31 @@ def freq_table(var_series, direction_series, var_bin_array=np.arange(-0.5, 41, 1
...
+
+    A missing data period is one where data is not available for some consecutive timestamps. This breaks
+    time continuity of the data. The function returns a dataframe where the first column is the starting timestamp
+    of the missing period and the second column is the end date of the missing period. An additional column also shows
+    how many days of data were lost in a missing period.
+
 
     :param data: Data for checking continuity, timestamp must be the index
     :type data: pandas.Series or pandas.DataFrame
     :return: A dataframe with days lost and the start and end date between them
```
stephenholleran: Rephrase the return comment to be:
A DataFrame with the start and end timestamps of missing gaps in the data along with the size of the gap in days lost.

brightwind/analyse/analyse.py
```diff
@@ -326,17 +326,31 @@ def freq_table(var_series, direction_series, var_bin_array=np.arange(-0.5, 41, 1
 
 def time_continuity_gaps(data):
     """
-    Returns the gaps in timestamps for the data, that means that data isn't available for that period.
+    Returns the start and end timestamps of missing data periods. Also days lost.
+
+    A missing data period is one where data is not available for some consecutive timestamps. This breaks
+    time continuity of the data. The function returns a dataframe where the first column is the starting timestamp
```
stephenholleran: Explaining how the function works should explain that the sample period is automatically calculated using '_xxxxx' and this is then used to scroll? through the time series to find gaps. The reason for explaining all of this is the user will be able to figure out what might be going wrong if they send in data that has an inconsistent sampling period?

Issue #15, opened 2019-01-14
[monthly_means] modify plot when monthly coverage is returned


stephenholleran:
When 'return_coverage=True' the plot should be modified to include bars indicating the monthly coverage. e.g. 
![image](https://user-images.githubusercontent.com/25622575/51108482-4c377600-17ea-11e9-9a5d-a0a15e5d015c.png)

This will obviously get very cluttered when multiple wind speeds are sent but that is up to the user to minimise what to send.

Try and match the colours in the example for the first wind speed column.

Inder95:
![image](https://user-images.githubusercontent.com/23189856/56520314-094d1300-653c-11e9-9363-ec620445d8a1.png)

Inder95:
![image](https://user-images.githubusercontent.com/23189856/56659343-c488bf80-6694-11e9-8cf5-16c82cd7cda1.png)
![image](https://user-images.githubusercontent.com/23189856/56659372-d79b8f80-6694-11e9-85ac-6b7242c23d45.png)
![image](https://user-images.githubusercontent.com/23189856/56659399-e7b36f00-6694-11e9-948c-208a3aa0d936.png)
![image](https://user-images.githubusercontent.com/23189856/56659418-f4d05e00-6694-11e9-9a05-3798427869fe.png)

Inder95:
@stephenholleran your suggestions in GH-41  implemented here.

stephenholleran:
Great.

Last thing. Add in these 2 lines to make it look like below. If that works then feel free to merge the branch.
```
ax2.yaxis.grid(True)
ax2.set_axisbelow(True)
```

![image](https://user-images.githubusercontent.com/25622575/56663245-8cd24580-669d-11e9-9345-6eeb23c509a3.png)

Inder95:
![image](https://user-images.githubusercontent.com/23189856/56673919-09baea80-66b1-11e9-955e-1cd3683e71f2.png)

PR #41, opened 2019-04-04
Iss15 add coverge to monthly means plot

Inder95:
@stephenholleran Added the coverage plot to monthly_means as described in iss #15. But if multiple series are passed it does not plot the coverage. I did because it would be make the graph useless because all the coverage bars would be at the same place on x-axis. Doing this would save the user from thinking that there is some problem with the code.


stephenholleran:
Do you have a screen shot of the resulting plot using just a single wind speed variable? And more than one.

Inder95:
For one variable

image

For more than one variable as I mentioned I avoided plotting coverage to prevent confusion

image

stephenholleran:
Looks good.

Can you put units in for coverage [%] and capital 'C' for coverage? We should multiply by 100 as a  result?

What about the label on the left y-axis? Can we guess what is sent or have a default parameter set to 'Wind speed [m/s]'?

stephenholleran:
As with the lower image we should possibly specify which anemometer is plotted e.g. Spd80mS. The image I pasted in #15 has the labels to the left but they could be overlaid like in image 2 here or put above the plot?

Inder95:
> Looks good.
> 
> Can you put units in for coverage [%] and capital 'C' for coverage? We should multiply by 100 as a result?
> 
> What about the label on the left y-axis? Can we guess what is sent or have a default parameter set to 'Wind speed [m/s]'?

We are using coverage_threshold also with the range 0-1 hence converting to 5 from 0 - 100 might be confusing here. Coverage is ratio so no units might be fine here.

Inder95:
@stephenholleran How does this look? I can resolve the  conflicting issues if you are fine with merging it.

image

stephenholleran:
Its good. A few more suggestions to make it more complete.

1. Label for coverage 'Coverage [-]'

2. Make the coverage axis have a min and max of 0 and 1. I have a plot where there is no 100% coverage but the bars go right up to the top making it look like they have 100% coverage. There should be a gap hence setting the max to 1 should create this gap.

3. Make the min wind speed label set to 0. So it is consistent from one plot to the next.

4. Is it possible to add white space to the left of the first bar and to the right of the last bar. It looks a bit weird when there are only a few months in the plot.

stephenholleran:
Also, your code in the screen shot above. That doesn't work for me and I wouldn't expect it to.

image

Not sure what `<class 'str'>` is?



This is how I would expect it to work and it does. How is yours working?

image

Inder95:
> Also, your code in the screen shot above. That doesn't work for me and I wouldn't expect it to.
> 
> image
> 
> Not sure what `<class 'str'>` is?
> 
> This is how I would expect it to work and it does. How is yours working?
> 
> image

That print statement was for me checking something, didn't push it after I removed it.

Inder95:
image

image

stephenholleran:
It looks like there is no space after coverage and before the brackets. 

The x axis labels are not showing whole months in the second plot?

How is your code line showing a plot when a plot and data are returned?

PR #49, opened 2019-05-01
Iss18 offset timestamps

Inder95:
@stephenholleran For you perusal. Thorough tests were added covering most use cases. Please check if the docstring is descriptive enough.


stephenholleran:
Good to go if you fix the conflicts.

Issue #53, opened 2019-05-08
[load_windographer] add new function to load wind data exported from Windographer


stephenholleran:
This function will be a wrapper around the load_csv function where kwrgs are set specific for the windographer format.

```python
# Windographer format. Implement in brightwind.
df = bw.load_csv(file, skiprows=12, delimiter='\t', encoding='cp1258')
# Put this in a tutorial for the library and (if needed) make it standard in load_windographer
df = df.replace(9999, np.NAN)
```

As with other load_xxxx functions it will return a DataFrame.

PR #60, opened 2019-05-10
Iss #53 load windographer txt

stephenholleran:
Add a function to load a Windographer formatted data file exported from the software.

@Inder95 can you take a look at this and merge if happy?

### DIFF
--- a/brightwind/load/__init__.py
+++ b/brightwind/load/__init__.py
@@ -1,1 +1,0 @@
-__all__ = ['load_timeseries', 'load_timeseries_from_multiple_files', 'load_brightdata']
--- a/brightwind/load/load.py
+++ b/brightwind/load/load.py
@@ -15,9 +15,8 @@
 #     along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import pandas as pd
-import numpy as np
 import requests
-from typing import List, Dict
+from typing import List
 import errno
 import os
 import shutil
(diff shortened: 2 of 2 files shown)


stephenholleran:
> * We can add a windographer txt file to our datasets too

I'll  modify the campbell scientific demo file
> * number_of_header_rows can be argument with default 12, because there is an option in windographer where you can change the header

It looks for 'Date/Time' to define the header so no need to send anything as an argument.
> * Error when a folder path is sent instead of a file path should be in is_file() because it is used by other functions too and would prevent duplication.

That is covered already. It checks if it is a file or folder. If a file then it runs. If a folder it raises an error saying that it is a folder and should be a Windographer .txt file.

Inder95:
`
def _is_file(file_or_folder):
    """
    Returns True is file_or_folder is a file.
    :param file_or_folder: The file or folder path.
    :type file_or_folder: str
    :return: True if a file.
    """
    if os.path.isfile(file_or_folder):
        return True
    elif os.path.isdir(file_or_folder):
        return False
    else:
        raise FileNotFoundError("File or folder doesn't seem to exist.")
`
What I mean is instead of returning False from the _is_file() when a folder is sent we raise an error
   `raise FileNotFoundError("File path seems to be a folder. Please load a single Windographer .txt data file.")` as you have done within the load_windographer_txt()

stephenholleran:
OK. Can you write up a new issue to tackle that directly and assign me to it? Cheers.

PR #69, opened 2019-05-20
Iss39 add calc air density

Inder95:


stephenholleran:
@Inder95 did I resolve the conflicts properly? If so, you can merge. Cheers.

Issue #57, opened 2019-05-09
[twelve_by_24] update doc string and write tests


stephenholleran:
Update the doc string:

- [x] return is not just a DataFrame but is default a plot with DataFrame optional.

- [x] change 'var_name' to 'var_name_label' as this is the label that is applied to the legend. Makes it more intuitive about what 'var_name' will actually do.

- [x] :parm and :type for 'var_series' isn't well defined.

Write tests:

- [x] ??

Inder95:
@stephenholleran For your perusal

stephenholleran:
> Accepts a variable series and returns a plot of 12x24 (months x hours) table for the variable.

Change to something like this:
Accepts a variable series and returns a plot of 12x24 (12 months x 24 hours) for the 'mean' of the variable with the table of data as an optional return. The aggregation_method 'mean' can be can be changed as outlined below.

- [ ] Keep the doc string parameters in the same order as they are listed in the function itself. Makes it easier for the user to follow when trying to figure out what each parameter is for. Maybe return_data as the last one to be somewhat consistent with other functions.

- [ ] Can this return a count or frequency as well as the percentile?

- [ ] In the example usage can you give an example for when a function is passed for the aggregation method? Maybe write a test for this too?

Inder95:
> > Accepts a variable series and returns a plot of 12x24 (months x hours) table for the variable.
> Change to something like this:
> Accepts a variable series and returns a plot of 12x24 (12 months x 24 hours) for the 'mean' of the variable with the table of data as an optional return. The aggregation_method 'mean' can be can be changed as outlined below.

Changed


> * [ ]  Keep the doc string parameters in the same order as they are listed in the function itself. Makes it easier for the user to follow when trying to figure out what each parameter is for. Maybe return_data as the last one to be somewhat consistent with other functions.

return_data moved to end as suggested

> * [ ]  Can this return a count or frequency as well as the percentile?

Can return count, added that to docstring but not frequency or percentile

> * [ ]  In the example usage can you give an example for when a function is passed for the aggregation method? Maybe write a test for this too?

Added

stephenholleran:
> > * [x]   Can this return a count or frequency as well as the percentile?
> 
> Can return count, added that to docstring but not frequency or percentile

To me count and frequency are the same thing.
Should we add percentile so? In other distribution functions we use '%frequency' to calculate percentages. What do you think?

Inder95:
[...]

PR #75, opened 2019-05-28
Add tests and update docstring for twelve_by_24(). Iss #57

Inder95:
Pull request for iss #57

### DIFF
--- a/brightwind/analyse/analyse.py
+++ b/brightwind/analyse/analyse.py
@@ -452,8 +452,8 @@ def twelve_by_24(var_series, aggregation_method='mean', var_name_label='', retur
     :param var_series: Variable to compute 12x24 for
     :type var_series: pandas.Series
     :param aggregation_method: 'mean' by default, calculates mean of the variable passed. Can change it to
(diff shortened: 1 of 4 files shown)


stephenholleran:
See comments in https://github.com/brightwind-dev/brightwind/issues/57#issuecomment-496443384

Inder95:
Implemented suggestions in https://github.com/brightwind-dev/brightwind/issues/57#issuecomment-498385996.

brightwind/analyse/analyse.py
```diff
@@ -479,19 +479,16 @@ def custom_agg(x):
         graph, table12x24 = bw.twelve_by_24(df.PrcpTot, aggregation_method=custom_agg, return_data=True)
 
     """
+    var_name_label = var_series.name if var_name_label == "" else var_name_label
```
Inder95: This won't work if series is sent in the form df[['T2m']] (as a dataframe) and by convention we will set the default of var_name_label as None instead of ''. Implemented these changes in the commit after this.

PR #114, opened 2019-06-23
Iss58 add scatter plots

stephenholleran:


stephenholleran:
@Inder95 after you have reviewed this and are happy could you update the live documentation website with what we have till now?

Issue #203, opened 2019-10-18
[SpeedSort] If synthesize with setup data, it fails


stephenholleran:
- [ ] `filter_by_coverage_threshold=False` doesn't exist in function `average_data_by_period`. Make sure that simply taking out this parameter is all that is needed to fix this.

- [ ] Why didn't a test pick this up?

stephenholleran:
Also, the synthesize method doesn't seem to work as expected and is doing something funny. Maybe it is linked to me simply removing the `filter_by_coverage_threshold` to get this to work.

```
data = bw.load_csv(bw.datasets.demo_data)
Spd80mN_mth_plt, Spd80mN_mth_data = bw.monthly_means(data.Spd80mN, return_coverage=True)
display(Spd80mN_mth_plt)
Spd80mS_mth_plt, Spd80mS_mth_data = bw.monthly_means(data.Spd80mS, return_coverage=True)
display(Spd80mS_mth_plt)
```

![image](https://user-images.githubusercontent.com/25622575/67089296-27b86200-f19f-11e9-99a1-c89a8b946b64.png)

stephenholleran:
```
ss_obj = bw.Correl.SpeedSort(data.Spd80mN, data.Dir78mS, data.Spd80mS, data.Dir78mS, '10min')
ss_obj.run(False)
data['Spd80mS_Synthesized'] = ss_obj.synthesize()['Spd80mS_Synthesized']
Spd80mS_Synth_mth_plt, Spd80mS_Synth_mth_data = bw.monthly_means(data.Spd80mS_Synthesized, return_coverage=True)
display(Spd80mS_Synth_mth_plt)
```
![image](https://user-images.githubusercontent.com/25622575/67089478-97c6e800-f19f-11e9-8dc1-5ba7129ee6aa.png)

stephenholleran:
The result of the synthesize method is strange. First Spd80mS_Synthesized is cut off in Aug sometime?

Also, even though there is Spd80mN data available up till Nov-2016, this data is not used to synthesize the missing data at Spd80mS?

lucunnin:
Spent a while looking into this. As far as I can tell, removing `filter_by_coverage_threshold` has no effect and is fine. 

[...]

PR #211, opened 2019-11-15
iss #203 remove filter_by_coverage_threshold

stephenholleran:


stephenholleran:
iss #203 remove filter_by_coverage_threshold

PR #217, opened 2019-11-22
iss214 fix return plot bug

lucunnin:
I have fixed a bug in the correlation code where some plots were not being returned properly.


stephenholleran:
Fixed in issue #261 and pull request #270

PR #218, opened 2019-11-22
iss215 make OrthogonalLeastSquares.linear_func a static function

lucunnin:
OrthogonalLeastSquares.linear_func has been changed to a static function, consistent with OrdinaryLeastSquares.linear_func.


stephenholleran:
Fixed in issue #261

Issue #282, opened 2021-02-04
[Speedsort]  Few bugs in synthetize function


BiancaMorandi:
- [x] the argument wdir_column_names=self._ref_dir_col_name is not given to tf.average_data_by_period in order to do a vector average for direction
![image](https://user-images.githubusercontent.com/65398058/106922032-855e1e00-6704-11eb-8fc6-4a242a76a9c6.png)

- [x] the function is crashing if the sector is present in the averaged ref_dir but not in the averaged ref_dir concurrent with target_dir. ie. the code below is giving the error in the screenshot below
veer = bw.Correl.SpeedSort(MERRA2_NE['WS50m_m/s']['2016-03-02 00:00:00':], 
                           MERRA2_NE['WD50m_deg']['2016-03-02 00:00:00':],
                           DATA_CLND['Spd80mN'][DATA_CLND.index.month.isin([2, 4, 6, 8, 10, 12])], 
                           DATA_CLND['Dir78mS'][DATA_CLND.index.month.isin([2, 4, 6, 8, 10, 12])], 
                           averaging_prd='1M')
veer.run(show_params = False)
veer.synthesize()
![image](https://user-images.githubusercontent.com/65398058/106922972-804d9e80-6705-11eb-9ff9-7187f9753853.png)

- [x] there is not a test for the Speedsort synthesize() function in test_correlation.py

BiancaMorandi:
Additional info for: _'the function is crashing if the sector is present in the averaged ref_dir but not in the averaged ref_dir concurrent with target_dir.'_
Need to check that _predict_dir is doing the same thing than vba code below.
![image](https://user-images.githubusercontent.com/65398058/107024482-03bbce00-67a0-11eb-97b7-0d5e2fe31996.png)
vba code
![image](https://user-images.githubusercontent.com/65398058/107024578-20580600-67a0-11eb-93df-ff12195664e9.png)

PR #286, opened 2021-02-23
Iss282 speed sort bug

BiancaMorandi:
Fixed all problems reported in #282 issue


brightwind/analyse/correlation.py
```diff
@@ -949,53 +952,85 @@ def plot(self):
...
+                prediction_spd = data['spd'] * np.nan
+
+            prediction = pd.concat([prediction, prediction_spd], axis=0)
 
         return prediction.sort_index()
 
     def synthesize(self, input_spd=None, input_dir=None):
         # This will give erroneous result when the averaging period is not a whole number such that ref and target does
-        # bot get aligned -Inder
+        # both get aligned -Inder
```
stephenholleran, line 1023: Is Inder's comment still valid since I refactored the average_data_by_period function?

tests/test_correlation.py
```diff
@@ -200,6 +200,33 @@ def test_synthesize():
...
+                                                 '2016-08-01': 7.093956,
+                                                 '2016-09-01': np.NaN,
+                                                 '2016-10-01': 6.669446,
+                                                 '2016-11-01': np.NaN,
+                                                 '2016-12-01': 8.900778,
+                                                 '2017-01-01': np.NaN,
+                                                 '2017-02-01': 9.134509,
+                                                 '2017-03-01': np.NaN},
+                         'Dir78mS_Synthesized': {'2016-03-01': 257.413204,
+                                                 '2016-04-01': np.NaN,
```
stephenholleran, line 217: The resulting synthesised data has wind speed for 20166-0-01 but there is no wind direction. How can there be wind speed if there is no wind direction?

brightwind/analyse/correlation.py
BiancaMorandi: solved

PR #318, opened 2022-01-17
Iss309 sector ratio subplot

rach185:
Added subplotting functionality.

### DIFF
--- a/brightwind/analyse/analyse.py
+++ b/brightwind/analyse/analyse.py
@@ -1354,7 +1354,10 @@ def sector_ratio(wspd_1, wspd_2, wdir, sectors=72, min_wspd=3, direction_bin_arr
         if len(sec_rats_dists) == 1:
             return fig, sec_rats_dists[0]
         else:
-            return fig, sec_rats_dists
+            sec_rats_df = pd.DataFrame(index=sec_rats_dists[0].index)
+            for key in sec_rats_dists:
+                sec_rats_df[key] = sec_rats_dists[key]
+            return fig, sec_rats_df
     return fig
 
 
--- a/brightwind/analyse/analyse.py
+++ b/brightwind/analyse/analyse.py
@@ -1129,10 +1129,14 @@ def sector_ratio(wspd_1, wspd_2, wdir, sectors=72, min_wspd=3, direction_bin_arr
     :type: min_wpd: float
     :param direction_bin_array: (Optional) Array of numbers where adjacent elements of array form a bin. This
                                  overwrites the sectors.
-    :param boom_dir_1: Boom direction in degrees of wspd_1. If top mounted leave default as -1.
-    :type boom_dir_1: float
-    :param boom_dir_2: Boom direction in degrees of wspd_2. If top mounted leave default as -1.
-    :type boom_dir_2: float
+    :param boom_dir_1: Boom direction in degrees of wspd_1. If top mounted leave default as -1. One or more boom
+                       orientations can be accepted. If multiple orientations, number of orientations must equal
+                       number of anemometer pairs.
+    :type boom_dir_1: float or list
+    :param boom_dir_2: Boom direction in degrees of wspd_2. If top mounted leave default as -1. One or more boom
+                       orientations can be accepted. If multiple orientations, number of orientations must equal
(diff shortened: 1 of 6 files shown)


brightwind/analyse/analyse.py
```diff
@@ -1156,25 +1161,80 @@ def sector_ratio(wspd_1, wspd_2, wdir, sectors=72, min_wspd=3, direction_bin_arr
...
+
+        sec_rat_plot, sec_rat_dist = dist_by_dir_sector(sec_rat.loc[common_idx], wdir.loc[common_idx], sectors=sectors,
+                                                        aggregation_method='mean',
+                                                        direction_bin_array=direction_bin_array,
+                                                        direction_bin_labels=None, return_data=True)
+        matplotlib.pyplot.close()
+        sec_rat_dist = sec_rat_dist.rename('Mean_Sector_Ratio').to_frame()
+        sec_rats_dists[sensor_pair] = sec_rat_dist
+
+        fig = plt.plot_sector_ratio(sec_ratio=sec_rats, wdir=wdir, sec_ratio_dist=sec_rats_dists, col_names=col_names,
```
BiancaMorandi, line 1228: we should allow to have multiple boom_dir_1 and boom_dir_2. As sensors at different heights can be mounted on booms with different orientation. Do you think that it would be easy to fix this?

brightwind/analyse/plot.py
```diff
@@ -987,38 +987,114 @@ def plot_12x24_contours(tab_12x24, label=('Variable', 'mean'), plot=None):
     return ax.get_figure()
```
BiancaMorandi, line 989: can you add a test for `plot_sector_ratio` function in test_plot.py?

BiancaMorandi (commented):
@rach185 I have finally reviewed this. Can you please address my comments when you have time? Happy to discuss

brightwind/analyse/analyse.py
```diff
@@ -1156,25 +1161,80 @@ def sector_ratio(wspd_1, wspd_2, wdir, sectors=72, min_wspd=3, direction_bin_arr
...
+        #To change the radius limits of plot and the figure size
+        bw.sector_ratio(data.Spd80mN, data.Spd80mS, wdir=data.Dir78mS, radial_limits=(0.8, 1.2), figure_size=(10,10))
+
+        #To create subplots with different anemometers
+        bw.sector_ratio(data[['Spd80mN', 'Spd60mN']], data[['Spd80mS', 'Spd60mS']], data['Dir78mS'],
+                        boom_dir_1=0, boom_dir_2=180, figure_size=(25,25))
+
+        #To use different wind vanes with each anemometer pair
+        bw.sector_ratio(data[['Spd80mN', 'Spd60mN', 'Spd40mN']], data[['Spd80mS', 'Spd60mS', 'Spd40mS']],
+                        data[['Dir78mS', 'Dir58mS', 'Dir38mS']], boom_dir_1=0, boom_dir_2=180, figure_size=(25,25))
```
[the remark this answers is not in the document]
stephenholleran: Can this be done by adding font size parameter to be sent to the function?

Or, I see that there are **kwargs allowed, you could give an example in the doc string to show what the parameter names are to change the font size?

stephenholleran:
@rach185 could you also add this to the CHANGELOG file
https://github.com/brightwind-dev/brightwind/blob/dev/CHANGELOG.md

brightwind/analyse/analyse.py
rach185: in analyse.sector_ratio

- Accept int or list of ints as input params
- if list: check len is equal to len wspd_1 (and wspd_2)

- Update doc string
- Add example

in plot.plot_sector_ratio

- if list: check len is equal to something sensible in sec_ratio input
- add a boom counter to iterate through if list: boom_count = 0
- for pair in sec_ratio: if boom_dir is a list, take first element in list boom_dir[0], continue through loop then increment boom_counter += 1 befor next itt. If boom_dir is an int then change nothing

- Update doc string

stephenholleran: @rach185 @BiancaMorandi this conversation would be better had in the actual issue as that is where things will be better tracked.

Issue #224, opened 2020-02-19
[plot_scatter_wdir] and [plot_scatter] plots 1 series against itself if 2 series have same name.


AndyBrightWind:
If 2 different series are passed to the function which both have the same name (column label), the returned plot shows a perfect correlation (only plotting one series against itself). If one of these names is changed before being passed to the function, the correct scatter is plotted. This issue doesn't occur for bw.Correl.OrdinaryLeastSquares() or pyplot.plot().

PR #322, opened 2022-01-20
Iss #224 plot scatter duplicate name

BiancaMorandi:
issue #224

fixed bug, now if x and y pandas series have same name then the correct scatter plot is generated


brightwind/analyse/plot.py
```diff
@@ -459,6 +459,11 @@ def plot_scatter(x, y, trendline_y=None, trendline_x=None, line_of_slope_1=False
     elif type(y) is np.ndarray or type(y) is list:
         y = pd.Series(y).rename('y')
 
+    # if x and y names are the same then rename pd.Series names to be unique
+    if x.name == y.name:
+        x = x.rename(x.name + '_0')
```
BiancaMorandi, line 464: change suffix to be _x and _y

Issue #324, opened 2022-03-04
Replace all 'is' and 'is not' with '==' or '!='


stephenholleran:
Python 3.8+ throws a SyntaxWarning if 'is' or 'is not' is found in the code. Reason is:

![image](https://user-images.githubusercontent.com/25622575/156817457-d424738a-764c-43c8-8250-f7e0f3b4ed34.png)

https://adamj.eu/tech/2020/01/21/why-does-python-3-8-syntaxwarning-for-is-literal/

stephenholleran:
There is no need to replace this when we are comparing to singularity classes such as `None` or `float`.
Therefore there is only 1 place to update this and even at that it compares to the value 1 which is ok to use. But I have updated it anyway.

BiancaMorandi:
merged

PR #328, opened 2022-03-25
iss #324 replace is with ==

stephenholleran:
iss #324


stephenholleran:
I have deliberately not included in the CHANGELOG as it is such a small change that has absolutely no impact apart from not throwing the warning.

Issue #307, opened 2021-10-11
[basic_stats] issue with numpy version 1.20.1


SteveCordleBW:
Running basic_stats gives an error - it still works with numpy version 1.18.1, but not with the latest version (1.20.1). Error shown below and in attached image.

Cannot interpret '<attribute 'dtype' of 'numpy.generic' objects>' as a data type

![image](https://user-images.githubusercontent.com/91064627/136812867-af98cac8-49ad-4b8b-9407-2ffce0e6e15a.png)

BiancaMorandi:
bug was not due to numpy version but probably to a bug of pandas when used with python version 3.8

PR #329, opened 2022-04-13
[basic_stats] iss #307 fixed bug due to pandas version and python 3.8

BiancaMorandi:
I think that python version used when having this error was Python 3.8.x, as I was able to replicate this issue ONLY when using Python 3.8.13. This error was raised with any `numpy version > 1.18.1` and `pandas==0.25.3`. This is probably due to a bug in ...\Anaconda3\envs\Python38\lib\site-packages\pandas\core\generic.py -> `describe` function

the `describe` function is taking as default input `include=None` but then np.numeric doesn't work. As for description below, for our application of describe function there is no need to give as input `include=None` but `include="All"` is ok and is not generating this error.

image

### DIFF
--- a/brightwind/analyse/analyse.py
+++ b/brightwind/analyse/analyse.py
@@ -907,9 +907,9 @@ def basic_stats(data):
 
     """
     if isinstance(data, pd.DataFrame):
-        return data.describe(percentiles=[0.5]).T.drop(['50%'], axis=1)
+        return data.describe(percentiles=[0.5], include='all').T.drop(['50%'], axis=1)
     else:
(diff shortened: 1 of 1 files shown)


BiancaMorandi:
I have tested this fix with versions below:

- Python 3.7.6 -> numpy==1.18.1 -> pandas==0.25.3
- Python 3.8.13 -> numpy==1.18.1 -> pandas==0.25.3
- Python 3.8.13 -> numpy==1.22.3-> pandas==0.25.3
- Python 3.9.12 -> numpy==1.22.3 -> pandas==1.3.2

PR #368, opened 2023-02-01
[Iss356] fix pandas version warnings

BiancaMorandi:
1. Addressed all warnings (Future and Deprecation) raised with Pandas<=1.5.3 and one error raised with Pandas>= 1.4.4
     1. Tests carried out with:
        - matplotlib >= 3.1.3, <=3.6.3
        - numpy >=1.18.1, <= 1.24.1
        - pandas >=0.25.3, <=1.5.3
        - Python >=3.7.6, <= 3.10.9
1. updated pandas requirements and changelog


stephenholleran:
Hi @BiancaMorandi,

> 1. Addressed all warnings (Future and Deprecation) raised with Pandas<=1.5.3 and one error raised with Pandas>= 1.4.4
>    
>    1. Tests carried out with:
>       
>       * matplotlib >= 3.1.3, <=3.6.3
>       * numpy >=1.18.1, <= 1.24.1
>       * pandas >=0.25.3, <=1.5.3
>       * Python >=3.7.6, <= 3.10.9
> 2. updated pandas requirements and changelog

In the requirements file we have:

```
pandas          >= 0.24.0, <=1.5.3
numpy           >= 1.16.4
matplotlib      >= 3.0.3
```
What is the situation for:
1. pandas 0.24.0 to 0.25.3, 
2. matplotlib 3.0.3 to 3.1.3 and 
3. numpy 1.16.4 to 1.18.1?

BiancaMorandi:
@stephenholleran I did some additional tests with versions below. Tests are failing only for pandas < 0.25.0, I have therefore updated the minimum pandas requirement to 0.25.0.
 
- pandas 0.24.0 to 0.25.3,
- matplotlib 3.0.3 to 3.1.3 and
- numpy 1.16.4 to 1.18.1?

brightwind/analyse/shear.py
```diff
@@ -273,14 +277,14 @@ def __init__(self, wspds, heights, min_speed=3, calc_method='power_law', by_mont
 
                     elif i == segments_per_day - 1:
                         start = str(start_times[i].time())
-                        end = str(start_times[0].time())
-                        time_wspds[i] = pd.DataFrame(anemometers_df).between_time(start, end, include_end=False)
+                        end = str((start_times[0] - resolution).time())
```
stephenholleran, line 280: @BiancaMorandi I wonder would there be a case where the resolution is incorrect?

BiancaMorandi: @stephenholleran yes, it could happen if timeseries has many data gaps but I thought that for shear it will be unlikely that this will happen (user will usually give as many data as possible and not very irregular data).  If we want to handle this then I need to add an option to have resolution as a user input to the function. 

It doesn't make much sense to me to have resolution as an input as this is only used to make sure that each segment of the day is not overlapping...it was done with `.between_time(start, end, include_end=False)` function without using `resolution`. This function will be updated to use different inputs and if I change the inputs to the latest version I cannot have compatibly with old versions of pandas.

stephenholleran: ok, thanks. Don't worry about it. As you said for shear the resolution should be well defined.

brightwind/analyse/analyse.py
```diff
@@ -434,7 +434,21 @@ def custom_agg(x):
         if i_dist == 0:
             distributions = distribution
         else:
-            distributions = pd.concat([distributions, distribution], axis=1)
+            # The below if statement is a workaround to be able to work with pandas >= 0.24.0, < 0.25.0, this should
+            # be replaced with distributions = pd.concat([distributions, distribution], axis=1)
+            # when these versions of pandas will not be supported anymore by brightwind library. This version of pandas
+            # doesn't allow to concatenate two pandas.Series with a CategoricalIndex if having a different index length
+            if len(distributions) >= len(distribution):
```
stephenholleran, line 441: @BiancaMorandi can you add a card so we will remember to depreciate this in the next breaking change version? Thanks.

BiancaMorandi: done #372

Issue #345, opened 2022-07-18
[dist_of_wind_speed] doesn't allow to use different COLOR_PALETTE than the BW default


BiancaMorandi:
dist_of_wind_speed function should be updated to take as input the colours - this is the case for current `master` but latest updates for `dev` don't allow this

BiancaMorandi:
@stephenholleran as discussed, if you prefer that no color input is given to distribution speeds and that color list can be changed only changing the `bw.analyse.plot.COLOR_PALETTE.color_list ` then this issue can be closed and all changes disregarded. Let me know your thoughts.

stephenholleran:
hi @BiancaMorandi 

Sorry, I am getting confused between all the PRs and issues.

OK, lets not allow the user to send colour direct in the function but they can change them in the colour palette.

The main problem was that the `_bar_subplot()` wasn't using the color_palette in the first place though? Is this fixed?

stephenholleran:
> hi @BiancaMorandi
> 
> Sorry, I am getting confused between all the PRs and issues.
> 
> OK, lets not allow the user to send colour direct in the function but they can change them in the colour palette.
> 
> The main problem was that the `_bar_subplot()` wasn't using the color_palette in the first place though? Is this fixed in a different PR?

The colors are changed by setting the ` bw.analyse.plot.COLOR_PALETTE.color_list` and not by setting `bw.analyse.plot.COLOR_PALETTE.primary`.

PR #375, opened 2023-02-07
[Iss345] fix color dist of wind speed

BiancaMorandi:
added `bar_colors `inputs to `dist`,  `_bar_subplot`, `plot_freq_distribution`, `dist_of_wind_speed`


stephenholleran:
@BiancaMorandi is this the PR that we decided that we shouldn't allow the user to send colours directly in the function call but they should edit the colour palette instead?

BiancaMorandi:
> @BiancaMorandi is this the PR that we decided that we shouldn't allow the user to send colours directly in the function call but they should edit the colour palette instead?

@stephenholleran as for comment I have added to #345, if you prefer that no color input is given to distribution speeds and that color list can be changed only changing the bw.analyse.plot.COLOR_PALETTE.color_list  then this issue can be closed and all changes disregarded. Let me know your thoughts.

Issue #377, opened 2023-02-09
[SpeedSort] SpeedSort sector_predict function should use two curves for fitting data


BiancaMorandi:
The SpeedSort sector_predict function is predicting all speed data using the slope and offset derived for speed > cut off.  The 'The SpeedSort, DynaSort and Scatter Wind Correlation Methods, Wind Engineering 29(3):217-242, Ciaran King, Brian Hurley, May 2005'. paper reports that two different fit lines should be used for predicting target data from the reference speed. 

-> reference data < cut off speed should be adjusted using the line passing through zero in plot below
-> reference data >= cut off speed should be adjusted using the slope and offset for line defined from cut off speed and not passing through zero

![image](https://user-images.githubusercontent.com/65398058/217885628-21842e7e-f79b-4b62-90a2-243279ab02e7.png)

BiancaMorandi:
fixed big, the SpeedSort sector_predict function is now fitting the reference data using the two fit lines.

PR #379, opened 2023-02-10
[Iss377] speedsort sector predict update

BiancaMorandi:
The `SpeedSort.speed_model.sector_predict` function was updated to consider two fit lines:

1. line passing through zero and the cut off reference speed -> used for reference speeds < cut off
2. best fit line of points above cut off reference speed -> used for speeds >= cut off


BiancaMorandi:
> This is for issue #377
> 
> Hi @BiancaMorandi ,
> 
> This is good. Is it possible to include a test to cover this portion of the prediction? Even if x was 0, you should get 0 out.
> 
> Thanks,

@stephenholleran I have added a test for one sector and when speed is lower than cutoff.

PR #380, opened 2023-02-10
[Iss376] plot shear time of day colors

BiancaMorandi:
The `plot_shear_time_of_day` function was updated to remove the hardcoded colors used when `plot_type` is 'line' of 'step'. The following enhancements are made:

- a new color map was added to COLOR_PALETTE, color_map_symmetric_diverging
- colors for `plot_shear_time_of_day` are now taken from COLOR_PALETTE, color_map_symmetric_diverging
- user can change the colors to use for the color map only redefining  COLOR_PALETTE.color_map_symmetric_diverging

### DIFF
--- a/brightwind/analyse/plot.py
+++ b/brightwind/analyse/plot.py
@@ -16,7 +16,7 @@ import re
 import six
 from colormap import rgb2hex, rgb2hls, hls2rgb
 from matplotlib.ticker import PercentFormatter
-from matplotlib.colors import ListedColormap, LinearSegmentedColormap
+from matplotlib.colors import ListedColormap, to_hex
 import warnings
 
 register_matplotlib_converters()
@@ -73,7 +73,14 @@ class _ColorPalette:
         _col_map_colors = [self.primary_95,  # lightest primary
                            self.primary,     # primary
                            self.primary_10]  # darkest primary
+
+        _col_map_symmetric_diverging_colors = [self.primary_10,  # dark green, 10% of primary
+                                               self.fifth,       # yellow'ish, rgb(242, 216, 105)
+                                               self.tertiary]    # red'ish, rgb(155, 43, 44)
+
         self._color_map = self._set_col_map(_col_map_colors)
+
+        self._col_map_symmetric_diverging = self._set_col_map(_col_map_symmetric_diverging_colors)
 
         self.color_list = [self.primary, self.secondary, self.tertiary, self.fourth, self.fifth, self.sixth,
                            self.seventh, self.eighth, self.ninth, self.tenth, self.eleventh, self.primary_35]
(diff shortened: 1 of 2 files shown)


brightwind/analyse/plot.py
```diff
@@ -89,14 +96,29 @@ def _set_col_map(col_map_colors):
...
+    @property
+    def color_map_symmetric(self):
+        return self._color_map_symmetric
+
     @color_map.setter
     def color_map(self, col_map_colors):
         self._color_map = self._set_col_map(col_map_colors)
 
+    @color_map.setter
+    def color_map_symmetric(self, col_map_colors):
```
stephenholleran, line 108: @BiancaMorandi 
I'm not sure this works. The `@color_map.setter` I think is a setter for the `color_map` property. It looks like you have another property `color_map_symmetric` that you want to set so it might need to be `@color_map_symmetric.setter`?

BiancaMorandi: solved

BiancaMorandi: solved

brightwind/analyse/plot.py
```diff
@@ -1855,40 +1877,46 @@ def plot_log_law(avg_slope, avg_intercept, wspds, heights, max_plot_height=None)
 
 
 def plot_shear_time_of_day(df, calc_method, plot_type='step'):
+    """
+    Function used by Shear.TimeOfDay for plotting the hourly shear for each calendar month or an average of all months.
+    The color map used for plotting the shear by time of day for each calendar month depends on the plot_type input:
+        1) if 'line' 'step' the COLOR_PALETTE.col_map_symmetric_diverging is used
```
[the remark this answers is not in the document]
BiancaMorandi: solved

brightwind/analyse/plot.py
```diff
@@ -1885,40 +1907,46 @@ def plot_log_law(avg_slope, avg_intercept, wspds, heights, max_plot_height=None)
...
-    if calc_method == 'log_law':
+    elif calc_method == 'log_law':
         label = 'Roughness Coefficient'
+    else:
+        label = calc_method
 
     if plot_type == '12x24':
         df.columns = np.arange(1, 13, 1)
         df.index = np.arange(0, 24, 1)
-        df[df.columns[::-1]]
```
stephenholleran, line 1914: @BiancaMorandi I don't know what this did but why did you take it out?

stephenholleran:
@BiancaMorandi I have made some edits, renaming the color map from symmetric to cyclical and then modifying the colors.
What do you think?

brightwind/analyse/plot.py
BiancaMorandi: @stephenholleran I took this out because I don't think that is doing anything (statement seems to have no effect).

BiancaMorandi:
> @BiancaMorandi I have made some edits, renaming the color map from symmetric to cyclical and then modifying the colors. What do you think?

@stephenholleran looks good! thanks for these fixes. I have run also a test with latest versions of pandas, python, matplotlib, numpy (that are in requirements) and all good.

PR #383, opened 2023-03-15
[Iss382] seas adj dist matrix

rach185:
Fix iss382 regarding `_get_dist_matrix_by_dir_sector`. Error was thrown when `var_series` contained a month with zero coverage.

The solution is to only drop months which fail the coverage threshold if they have coverage greater than zero. If they have coverage equal to zero, then the month does not exist in the series.

Changed 
`var_series = tmp_var_series.set_index('Months').drop(labels=list( months_fail_coverage.index.strftime('%Y-%m'))).set_index(index_name).iloc[:, 0]`

To
`var_series = tmp_var_series.set_index('Months').drop(labels=list( months_fail_coverage[months_fail_coverage>0].index.strftime('%Y-%m'))).set_index(index_name).iloc[:, 0]`


BiancaMorandi:
@rach185 for tests it is better to assert against expected results instead of having only `assert True`. I have therefore updated the test and also made a minor change because code was generation a future warning in python=3.10.

@stephenholleran this issue is ready for your review.

PR #395, opened 2023-05-24
[Iss385] `slice_data`, `offset_timestamps`, `_LoadBWPlatform.get_data` date_to not included

BiancaMorandi:
I have updated `slice_data`, `offset_timestamps`, `_LoadBWPlatform.get_data` functions to consider date_to input as not included. e.g if date_to = 2023-02-01 then sliced data, offsetted data and data we get from platform are considered up to 2023-01-31 23:50. This is consistent with method used for `bw.LoadBrightHub.get_data` and `load_cleaning_file`.

As part of this branch I have also:

1. updated the docstrings and tests of `slice_data`, `offset_timestamps`, `_LoadBWPlatform.get_data`. And also doctsrings of `momm`, `plot_timeseries` functions as using  `slice_data`.
2. fixed bugs below for `offset_timestamps`:
   - In docstring it is stated that function can accept a datetime.time input but an error was raised when using this. This is fixed now.
   - In docstring it is stated that the output of the function is in same format as the input, this was not the case for datetime.datetime and datetime.date inputs. This is fixed now.

### DIFF
--- a/brightwind/transform/transform.py
+++ b/brightwind/transform/transform.py
@@ -1270,8 +1270,9 @@ def offset_timestamps(data, offset, date_from=None, date_to=None, overwrite=Fals
     """
     Offset timestamps by a certain time period
 
-    :param data:        DateTimeIndex or Series/DataFrame with DateTimeIndex
-    :type data:         pandas.DateTimeIndex, pandas.Series, pandas.DataFrame
+    :param data:        The timestamp or DateTimeIndex or Dataframe or Series to which apply the time offset.
+    :type data:         pandas.DateTimeIndex, pandas.Series, pandas.DataFrame, pandas.Timestamp, datetime.datetime,
+                        datetime.date, datetime.time
     :param offset:      A string specifying the time to offset the time-series.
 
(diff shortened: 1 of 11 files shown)


brightwind/transform/transform.py
```diff
@@ -1270,89 +1270,119 @@ def offset_timestamps(data, offset, date_from=None, date_to=None, overwrite=Fals
...
+    else:
+        date_from = pd.to_datetime(date_from)
+
+    if pd.isnull(date_to):
+        if isinstance(data, pd.DatetimeIndex):
+            date_to = data[-1]
+        elif isinstance(data, pd.Series) or isinstance(data, pd.DataFrame):
+            date_to = data.index[-1]
+    else:
+        date_to = pd.to_datetime(date_to) - datetime.timedelta(seconds=1)
```
stephenholleran, line 1358: Hi @BiancaMorandi ,
Can you remember why we take 1 sec off here but not in other places when we go to less than the date_to?

BiancaMorandi: @stephenholleran as requested I have updated `offset_timestamps` and `slice_data` functions to use less than for date_to instead of taking 1 sec off.

Issue #419, opened 2023-10-18
[load.BrightHub.get_data()] update url to pull from assembled file


stephenholleran:
**Is your feature request related to a problem? Please describe.**
The bw.BrightHub.get_data() underlying URL is changing.

![image](https://github.com/brightwind-dev/brightwind/assets/25622575/a00f555e-1056-4158-bc4a-6e80a3218c4a)

It is changing to: `"/measurement-locations/{}/timeseries-data"`.
At the moment there are no query parameters but there will be in the future for select a date range.
At the moment the whole timeseries data is returned. This removes the need to chunk the data as is the case at the moment. The speed to return will be greatly increased as the data will already be assembled. Previously, the timeseries data is assembled on the fly from the raw logger files.

**Describe the solution you'd like**
Suggested code:

```python
import requests
import pandas as pd
from io import StringIO
import numpy as np

def get_timeseries_data(measurement_station_uuid):
    """
    Get the timeseries data from BrightHub for a particular measurement station.

    :param measurement_station_uuid: A specific measurement station's uuid.
    :type measurement_station_uuid:  str
    :return:                         The timeseries data.
    :rtype:                          pd.DataFrame
    """
    response = bw.LoadBrightHub._brighthub_request(url_end="/measurement-locations/{}/timeseries-data"
                                                   .format(measurement_station_uuid))
    response_json = response.json()
    if 'Error' in response_json:  # catch if error comes back e.g. measurement_location_uuid isn't found
        raise ValueError(response_json['Error'])
    
    pre_signed_url = response_json["url"]
    timeseries_response = requests.get(pre_signed_url)
    return pd.read_csv(StringIO(timeseries_response.text,))
```

The timestamp column should be made the index before returning.

stephenholleran:
Merged.

PR #432, opened 2024-04-23
Iss419 brighthub get data update url

stephenholleran:
Please refer to issue #419.

### DIFF
--- a/brightwind/load/load.py
+++ b/brightwind/load/load.py
@@ -1310,6 +1310,27 @@ class LoadBrightHub:
     __DATE_FORMAT = "%Y-%m-%dT%H:%M:%S"
 
     @staticmethod
+    def get_timeseries_data(measurement_station_uuid):
+        """
+        Get the timeseries data from BrightHub for a particular measurement station.
+
+        :param measurement_station_uuid: A specific measurement station's uuid.
+        :type measurement_station_uuid:  str
+        :return:                         The timeseries data.
+        :rtype:                          pd.DataFrame
+        """
+        response = LoadBrightHub._brighthub_request(url_end="/measurement-locations/{}/timeseries-data"
+                                                    .format(measurement_station_uuid))
+        response_json = response.json()
+        if 'error' in response_json:  # catch if error comes back e.g. measurement_location_uuid isn't found
+            raise ValueError(response_json['error'])
+
+        pre_signed_url = response_json["url"]
+        timeseries_response = requests.get(pre_signed_url)
+        df = pd.read_csv(StringIO(timeseries_response.text, ))
+        return df.set_index('Timestamp')
+
+    @staticmethod
     def __get_date_range_as_chunks(date_from, date_to, interval):
         # Based on: https://stackoverflow.com/a/29721341/7373512
         diff = (date_to - date_from) / interval
--- a/brightwind/load/load.py
+++ b/brightwind/load/load.py
@@ -1312,11 +1312,11 @@ class LoadBrightHub:
     @staticmethod
     def __get_timeseries_data(measurement_station_uuid, date_from=None, date_to=None):
         """
-        Sub function to get the actual timeseries so retry and back off can be implemented.
+        Sub function to return the Brighthub GET timeseries-data API response.
         """
-        return LoadBrightHub._brighthub_request(url_end="/measurement-locations/{}/timeseries-data"
-                                                .format(measurement_station_uuid),
-                                                params={"date_from": date_from, "date_to": date_to})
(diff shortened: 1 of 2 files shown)


brightwind/load/load.py
```diff
@@ -1357,70 +1361,35 @@ def get_data(measurement_station_uuid, date_from=None, date_to=None):
...
+                raise ValueError("{}. {}".format(response_json['error'], response_json['details']))
+
+        if response.status_code == 404 and 'details' in response_json:  # other BrightHub specific errors.
+            raise ValueError("{}. {}".format(response_json.get('error', ''), response_json['details']))
+        elif 'error' in response_json:
+            raise ValueError("Unexpected error: Status Code {}. {}".format(response.status_code, response.text))
+
+        pre_signed_url = response_json["url"]
+        timeseries_response = requests.get(pre_signed_url)
+        df = pd.read_csv(StringIO(timeseries_response.text, ))
```
dancasey-ie, line 1391: remove empty second property

brightwind/load/load.py
```diff
@@ -1357,70 +1361,35 @@ def get_data(measurement_station_uuid, date_from=None, date_to=None):
...
+            # raise the 503 error if all the attempts have been used up
+            if response.status_code == 503 and assembly_in_progress_err_msg in response_json.get('details'):
+                raise ValueError("{}. {}".format(response_json['error'], response_json['details']))
+
+        if response.status_code == 404 and 'details' in response_json:  # other BrightHub specific errors.
+            raise ValueError("{}. {}".format(response_json.get('error', ''), response_json['details']))
+        elif 'error' in response_json:
+            raise ValueError("Unexpected error: Status Code {}. {}".format(response.status_code, response.text))
+
+        pre_signed_url = response_json["url"]
```
dancasey-ie, line 1389: call presigned_url instead of pre_signed_url to match BH code

brightwind/load/load.py
```diff
@@ -1357,70 +1361,35 @@ def get_data(measurement_station_uuid, date_from=None, date_to=None):
...
+        timeseries_response = requests.get(pre_signed_url)
```
dancasey-ie, line 1390: Consider error handling presigned_url request.

But the only one that comes to mind is if the presigned_url has expired but that should not happen when running this code. Or of there is an internet error

dancasey-ie (changes requested):

brightwind/load/load.py
stephenholleran: I think you put that in in the old chunking code. See line 1416 of what was deleted. I'll try it without it.

stephenholleran:
Thanks @dancasey-ie.

rach185:
I'm now testing this from bhdev. The date format that is expected isn't what i would expect, do you mean for the user to input yyyy-mm-ddThh:mm:ss? 

image

See the change to `date_from` which allows me to pull data.
image

stephenholleran:
> do you mean for the user to input yyyy-mm-ddThh:mm:ss?
No I don't.
It should work the same as all other brightwind functions where the date could be "YYYY-MM-DD" or "YYYY-MM-DD hh:mm:ss". We might need to convert the date a user puts in to the one with the "T" or else we should ask BrightHub to change. @dancasey-ie thoughts?

Also, you can call it just using ```bw.LoadBrightHub.get_data()```

dancasey-ie:
the brighthub GET /measurements-locations/{uuid}/timeseries-data API is set up to accept dates in the format  "YYYY-MM-DD hh:mm:ss" (no T) or "YYYY-MM-DDThh:mm:ss" (with T). 

Maybe brighthub should allow for dates with no time component, but it is not Brighthub convention so I do not think we will have in this release. @stephenholleran unless you think its a priority.

Alternatively, convert date strings to datetime strings using the datetime library before calling the brighthub api inside the brightwind library

stephenholleran:
Thanks @dancasey-ie.

@rach185 I think the best approach is to convert the date into a datetime within the brightwind library using the datetime library. So "_2016-04-01_" would become "_2016-04-01 00:00:00_". This could be a sub function so can be used else where. Are you able to implement this?

Thanks,

rach185:
I've created a subfunction, `__reformat_datetime_str`, in the LoadBrightHub class and updated the `__get_timeseries_data` function to use this.

stephenholleran:
Thanks @rach185.
I just did a bit of refactoring on it. Moved it up above the calling function and renamed to to say what it does.

Are you happy with everything else? If you are we'll merge.

rach185:
Thanks @stephenholleran. Yep I'm happy with everything else!

PR #440, opened 2024-06-17
Iss439 lidar filters

ShaneBrightWind:
Addressed issues of updating lidar functions for data model.


BiancaMorandi:
@ShaneBrightWind I had a quick look at the functions and the `remote_sensor_quality_filter` will work only for Windcube LiDARs as it seems to filter only based on availability variable. If we want this function to be generic for all RSD then we should be able to filter also based on SNR (for AQS Sodars) and Quality (for Triton Sodars) variables. Ideally these functions should be a bit more generic and do not depend on the filter measurement type name.

Issue #441, opened 2024-06-21
[Shear.TimeOfDay] Column labelling in shear coefficient matrix


rach185:
The shear by time of day method outputs a matrix of shear coefficients. The columns are labelled by months of the year `%b` which are incorrect. They should always be Jan-Dec but are sometimes May-Apr. The correct shear coefficient is applied when shearing a timeseries.

Alpha values replicated from a sheared from and shear to timeseries:
![image](https://github.com/brightwind-dev/brightwind/assets/69606609/149b6a1a-9af1-4fc3-a597-90f5495c3f31)

Alpha values output from `shear_obj.alpha`. These are displayed with May-Apr columns while it should be 1-12:
![image](https://github.com/brightwind-dev/brightwind/assets/69606609/7eb355f5-d431-4d2d-855a-da3236a93885)

Checks of individual timeseries values have been done to confirm this.

rach185:
When running this function of data, such that not every calendar month has adequate coverage, there is an error thrown. We should output a more useful error explaining the issue to the user.

BiancaMorandi:
@stephenholleran merge request for this issue is ready for you to review/approve. As part of this issue we have addressed the below:

- fixed bug where `shear.TimeOfDay` was not correctly applied if the input dataset starts from a month different than January and `by_month=True`
- fixed bug that the fix above had generated for when `by_month=False`
- raise more informative error when  `shear.TimeOfDay` is derived using less than a year of data but it is applied to a dataset with more than a year of data (e.g the error is stating that shear cannot be applied as there is missing month on shear object). See example code below.
- fixed bug for `plot_shear_time_of_day` where incorrect colors for each month were assigned if shear values are not present for one or more months
- added tests
- fixed tests for error text update that was made on #205 but not fixed on the tests
- updated changelog

```
data = bw.load_csv(bw.demo_datasets.demo_data)

anemometers = data[['Spd80mN', 'Spd60mN', 'Spd40mN']]
df = anemometers.copy()
df.loc[df.index.month >6, "Spd80mN"]  = 2
anemometers = [df["Spd80mN"], data['Spd60mN'], data['Spd40mN']]
# Specify the heights of these anemometers
heights = [80, 60, 40]

# Test initialisation
shear_by_tod_power_law3 = bw.Shear.TimeOfDay(anemometers, heights,calc_method='power_law', by_month = True, plot_type='line')
shear_by_tod_power_law3.alpha

[...]

Issue #205, opened 2019-10-18
[Shear] if not enough data, raise more informative error


stephenholleran:
If not enough concurrent data to calculate a shear for any of the functions (maybe even for a direction sector) the code throws an `IndexError: index 0 is out of bounds for axis 0 with size 0` error.
Capture this and raise an error with a more informative explanation.

```
data = bw.load_csv(bw.datasets.demo_data)
data = bw.apply_cleaning(data, r'C:\Users\Stephen\Documents\Analysis\demo_cleaning_file.csv')

# there is no data from Oct-2017 onwards for Spd80mS
Spd80mS_mth_plt, Spd80mS_mth_data = bw.monthly_means(data.Spd80mS, return_coverage=True)
Spd80mS_mth_plt
bw.Shear.Average([data.Spd80mS['2017-10-01':], data.Spd40mS], [80, 40])
```

BiancaMorandi:
this error should be raised if any of input data is nan or < min_speed over the full period.

BiancaMorandi:
@stephenholleran merge request for this issue is ready for you to review/approve. As part of this issue we have addressed the below:

- raise more informative error when input data is nan or < min_speed over the full period.
- `Shear.TimeSeries` and `Shear.TimeOfDay` has been updated to return np.nan when there is no valid data (e.g nan or < min_speed) for that timestamp or hour - see example code below
- added tests
- updated changelog

```
data = bw.load_csv(bw.demo_datasets.demo_data)
# # Data for a hour is 2 < 3 m/s
# df = data.copy()
# df.loc[df.index.hour == 2, "Spd80mN"]  = 2
# wspds_shear = [df["Spd80mN"], data.Spd40mS]
# heights_shear = [80, 40]
# wdir_shear = data.Dir78mS

bw.Shear.TimeOfDay(wspds_shear, heights_shear, calc_method='power_law')

bw.Shear.TimeSeries(wspds_shear, heights_shear)
obj_shear.alpha.loc[obj_shear.alpha.index.hour == 2].dropna()
```

FYI @olivia-bentley

PR #442, opened 2024-06-28
[Iss 441] Shear by Time of Day bug in _apply()

rach185:
The `bw.Shear._apply()` function uses integer months to index the shear coefficients. The shear coefficients matrix is labelled with calendar months and begins on the first full month of coverage in the input timeseries, i.e. does not always have the first column as January. This means that the shear coefficients applied to a particular month are not necessarily consistent. 

Currently the  `bw.Shear._apply()` function is updated to reference the coefficient matrix using calendar month names. Alternatively the coefficient matrix could be re-ordered to always begin in January - although may not be applicable for timeseries shorter than 1 year. 

Sensitivity checks on this fix are within the range of a 0.00%-0.13% impact on wind speeds shear up 60m, with the average impact being 0.04% change in sheared wind speed. However, these sensitivity checks were not extensive.


stephenholleran:
Hi @rach185,

I think we should get to the source of the problem which seems to be
image

we should order this resulting list so the first item is 1 (if there is a full year of data).

Then, we could use the same method for pulling the unique list of months instead of just 0 to 12 counting.
image

This should hopefully fix the issue?

rach185:
I have updated the `apply` function so that it iterates through the sorted list of unique months, I've enumerated through this list to keep the `j` variable and preserve how `df_wspds` is called.

rach185:
Reported bug when the coverage is too low in a month to calculate shear: column created for shear values but they cannot be filled -> conclusion, shear by TOD can only be used when all calendar months have good enough coverage. Error handling will be included.

BiancaMorandi:
> Reported bug when the coverage is too low in a month to calculate shear: column created for shear values but they cannot be filled -> conclusion, shear by TOD can only be used when all calendar months have good enough coverage. Error handling will be included.

@stephenholleran I see this comment added by Rachel on this issue and I wonder if we have added a bug on the fix we did for #205. After merging #205 into `dev` we are adding a nan for any hour where we don't have any data, e.g 1st table below (where data is missing for 2am). This means that when we apply the shear then any data for that missing hour is nan. So the overall coverage of the sheared data is lower than the original data. Is this what we want or we should raise an error instead as we want shear by Time of Day to be applied only when we have shear values for all hours and months?

image

BiancaMorandi:
merge request ready for @stephenholleran review/approval. Updates made as detailed on iss #441 .

stephenholleran (approved):

PR #448, opened 2024-11-05
[Iss 305] Added export_tws_file to export.py, and relevant tests

Ilirmc:
Added function to export climatology as WindSim format .tws file (export_tws_file()) and relevant tests.

### DIFF
--- a/brightwind/export/export.py
+++ b/brightwind/export/export.py
@@ -4,7 +4,7 @@ import datetime
 import re
 import brightwind
 
-__all__ = ['export_tab_file', 'export_csv']
+__all__ = ['export_tab_file', 'export_csv', 'export_tws_file']
 
 
 def _calc_mean_speed_of_freq_tab(freq_tab):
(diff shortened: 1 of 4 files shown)


brightwind/export/export.py
```diff
@@ -154,3 +154,97 @@ def export_csv(data, file_name=None, folder_path=None, **kwargs):
     else:
         raise NotADirectoryError("The destination folder doesn't seem to exist.")
     print('Export to csv successful.')
+
+
+def export_tws_file(eastings, northings, height, wspd_series, direction_series, wspd_std_series=None, site_name=None, file_name=None, folder_path=None):
+    """Export a WindSim timeseries tws file using a timeseries dataframe of wind speed and direction.
```
rach185, line 160: The description mentions 'timeseries dataframe of wind speed...' but you input a series, suggest removing 'dataframe'.

brightwind/export/export.py
```diff
@@ -154,3 +154,97 @@ def export_csv(data, file_name=None, folder_path=None, **kwargs):
...
+    :type wspd_std_series: pandas.Series
+    :param site_name: The site name to include in the file.
+           i.e 'Demo Data'
+    :type site_name: str
+    :param file_name: The file name under which the tab file will be saved, or use the default,
+        i.e '2019-06-07_brightwind_tws_export.tab'
+    :type file_name: str
+    :param folder_path: The directory where the tab file will be saved, default is the working directory.
+    :type folder_path: str
+
```
rach185, line 182: Add full stops to the ends of the sentences for param (it's bw library convention).

brightwind/export/export.py
```diff
@@ -154,3 +154,97 @@ def export_csv(data, file_name=None, folder_path=None, **kwargs):
...
+    :param height: Height that the timeseries represents in meters.
+    :type height: float
+    :param wspd_series: Series of wind speed variable
+    :type wspd_series: pandas.Series
+    :param direction_series: Series of wind directions between [0-360]
+    :type direction_series: pandas.Series
+    :param wspd_std_series: Series of wind speed standard deviations
+    :type wspd_std_series: pandas.Series
+    :param site_name: The site name to include in the file.
+           i.e 'Demo Data'
```
rach185, line 175: When defining `param site_name`, state that the default value will be `brightwind_site`.

rach185, line 175: The i.e. that you've included should align with the beginning of the sentence above (it's a bw library convention).

rach185: And again below.

brightwind/export/export.py
```diff
@@ -154,3 +154,97 @@ def export_csv(data, file_name=None, folder_path=None, **kwargs):
...
+    :type folder_path: str
+
+    :return: Creates a WindSim timeseries tws file which can be used in the WindSim software.
+
+    **Example Usage**
+    ::
+        import brightwind as bw
+        df = bw.load_campbell_scientific(bw.demo_datasets.demo_campbell_scientific_site_data)
+
+        bw.export_tws_file(df.Spd80mN, df.Dir78mS, 'Demo Data', 80, file_name='campbell_tws_file', 626100, 827971, folder_path=r'C:\some\folder\')"""
```
rach185, line 190: A few comments on this example:
1. Reorder the inputs here to reflect the function.
2. Assign `site_name='Demo Data'`.
3. Remove the folder path as the example will crash if a user runs it because it doesn't exist for them. It will also crash because you've added a trailing `\` which is interpreted as the string not finishing. You can add an example folder path with an `i.e.` above, like you've done before.

brightwind/export/export.py
```diff
@@ -154,3 +154,97 @@ def export_csv(data, file_name=None, folder_path=None, **kwargs):
...
+    
+    if site_name is None:
```
rach185, line 192: These 4 if statements can be written in one line the same way you've done it for `include_sd`

brightwind/export/export.py
```diff
@@ -154,3 +154,97 @@ def export_csv(data, file_name=None, folder_path=None, **kwargs):
...
+
+    if '.tws' not in file_name:
+        file_name = file_name + '.tws'
+
+    include_sd = True if wspd_std_series is not None else False
+    
+    file_name_print = os.path.splitext(file_name)[0]
+    file_path = os.path.join(folder_path, file_name)
+
+    eastings, northings = int(eastings), int(northings)
```
rach185, line 209: `int()` will always round down. As a user, I'd expect to round to the nearest whole number - unless WindSim requires otherwise? You can use `int(round())` instead.

brightwind/export/export.py
```diff
@@ -154,3 +154,97 @@ def export_csv(data, file_name=None, folder_path=None, **kwargs):
...
+    local_direction_series = brightwind.analyse.analyse._convert_df_to_series(direction_series.copy())
+    if include_sd: local_wspd_sd_series = brightwind.analyse.analyse._convert_df_to_series(wspd_std_series.copy())
+
+    current_timestamp = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    version = brightwind.__version__
+    
+    tws_string = f"{str(file_name_print)} created using brightwind version {version} at {current_timestamp}.\n"
+
+    tws_output = pd.concat([round(local_direction_series,4),round(local_wspd_series,4)],axis=1, keys=['dir:','speed:']).dropna()
+    if include_sd: tws_output = pd.concat([round(local_direction_series,4),round(local_wspd_series,4), round(local_wspd_sd_series,4)],axis=1, keys=['dir:','speed:','SDspeed:']).dropna()
```
rach185, line 221: As `tws_output` is already created you can add the stddev series to it, instead of overwriting it.

brightwind/export/export.py
```diff
@@ -154,3 +154,97 @@ def export_csv(data, file_name=None, folder_path=None, **kwargs):
...
+    :type file_name: str
+    :param folder_path: The directory where the tab file will be saved, default is the working directory.
+    :type folder_path: str
+
+    :return: Creates a WindSim timeseries tws file which can be used in the WindSim software.
+
+    **Example Usage**
+    ::
+        import brightwind as bw
+        df = bw.load_campbell_scientific(bw.demo_datasets.demo_campbell_scientific_site_data)
```
rach185, line 188: The load function crashed for me because the function should `bw.demo_datasets.demo_campbell_scientific_data` instead.

brightwind/export/export.py
```diff
@@ -154,3 +154,97 @@ def export_csv(data, file_name=None, folder_path=None, **kwargs):
...
+    if site_name is None:
+        site_name = 'brightwind_site'
+    
+    if file_name is None:
+        file_name = str(datetime.datetime.now().strftime("%Y-%m-%d")) + '_brightwind_tws_export.tws'
+
+    if folder_path is None:
+        folder_path = os.getcwd()
+
+    if '.tws' not in file_name:
```
rach185, line 201: I would suggest checking the last 4 char in the str being `.tws` as opposed to just checking if it comes up anywhere in the str.

brightwind/export/export.py
rach185: Additionally, how easy is WindSim to debug? Should you add in error checking to ensure the user has input the coords and heights as int instead of something else?

rach185 (commented):
@BiancaMorandi I have done a review of Ilir's code, once these are addressed, it should be ready for your review/approval!

brightwind/export/export.py
Ilirmc: Up[dated, however this is not the case in the export_tab_file?

brightwind/export/export.py
Ilirmc: Updated.

brightwind/export/export.py
Ilirmc: Updated.

brightwind/export/export.py
Ilirmc: Updated.

brightwind/export/export.py
Ilirmc: Updated, but this appears to also be incorrect in the export_tab_file() function.

brightwind/export/export.py
Ilirmc: Updated, but note that point 3 appears to also be incorrect in the export_tab_file() function.

brightwind/export/export.py
Ilirmc: This logic was used to be consistent with export_tab_file

Ilirmc: Updated.

brightwind/export/export.py
Ilirmc: Updated.

stephenholleran:
Hi @Ilirmc, @BiancaMorandi,

This we merged directly into the master branch!! PRs are supposed to go into the dev branch.

Also, I thought I was supposed to be the one to do the final review?

stephenholleran:
This didn't end up going into master with a closer look but what has gone on with this?

BiancaMorandi:
branch no merged to master and pull request closed without doing anything with it. We should ignore this request.

Issue #421, opened 2023-11-10
[Stations.measurements.get_table()] Pandas FutureWarning for `df.drop('meas_type_rank', 1, inplace=True)` not caught


stephenholleran:
**Describe the bug**
A FutureWarning in Pandas 1.5.3 was not accounted for to allow the library to work with newer Pandas versions. The FutureWarning was
![image](https://github.com/brightwind-dev/brightwind/assets/25622575/98427af8-7414-416a-aaed-c77aa46d78a7)

**To Reproduce**
Create a Station object and show the measurements table using Pandas==1.5.3.

**Expected behavior**
Plotting the table with no warning or error.

**Additional context**
The fix is to add the axis label.
```python
df.drop('meas_type_rank', axis=1, inplace=True)
```

BiancaMorandi:
this issue was not captured because there is not test using option that will trigger this. As part of this issue we need to add also a test e.g the example given above.

olivia-bentley:
new comment from Olivia

BiancaMorandi:
@stephenholleran merge request for this issue is ready for your review/approval. We have made the following changes:

-  fixed the bug due to pandas version, and run tests with v1.5.3 and 2.0.1
- removed the dropping of columns if all values are nan for `detailed=False`, `detailed=True`, `wind_speeds =True` and `wind_directions=True` input options. Hardcoded columns for each option are now shown on table even if all values are None but NOT shown if the dictionary of property is empty (e.g if a station doesn't have any settings for mounting_arrangements then relative columns are not shown on table)
- updated docstring of measurements.get_table()
- added tests to capture all user input options and examples given on docstring (added tests also for logger_measurement_configs .get_table() and MeasurementStation.get_table()
- updated gitignore file to ignore temp folder generated when running tests using vscode

PR #471, opened 2025-03-18
Iss421 pandas version bug fix

olivia-bentley:
Added the revised code and test for this for checking if pandas version is negatively influencing implementation of the code


brightwind/load/station.py
```diff
@@ -972,7 +972,7 @@ def get_table(self, detailed=False, wind_speeds=False, wind_directions=False, ca
             order_index = dict(zip(MEAS_TYPE_ORDER, range(len(MEAS_TYPE_ORDER))))
```
BiancaMorandi, line 972: can you update the doc string of the get_table function? as you can see on screenshot below this is not reporting the correct description for most of the parameters
image

brightwind/load/station.py
```diff
@@ -1039,7 +1039,7 @@ def get_table(self, detailed=False, wind_speeds=False, wind_directions=False, ca
             # get what is common from both lists and use this to filter df
```
BiancaMorandi, line 1039: it looks like a column is dropped if content is null for all rows, do you know why this happens? table should report all columns as for cols_required input even if content is null. I expect that this issue happens for all get_table() options. Is this a quick fix that can be handled in this branch? If not quick fix then I can generate a separate issue for this.

see example below where logger offset is not reported on table because all values are null

```
mm1 = bw.MeasurementStation(bw.demo_datasets.iea43_wra_data_model_v1_0)
mm1.measurements.get_table(wind_directions=True)
```


image

BiancaMorandi (changes requested):
@olivia-bentley see few comments. Happy to discuss.

brightwind/load/station.py
stephenholleran: @BiancaMorandi 
That seems like a Pandas default behaviour. If reasonably easy to do, we should probably display the column name even if all entries are empty.

tests/test_station.py
```diff
@@ -66,3 +66,132 @@ def test_get_table():
         if value is not '-':
             assert value in mm2.measurements.properties[4].values()
 
+def test_get_table_drop_methodology(monkeypatch):
```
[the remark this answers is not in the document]
olivia-bentley: this was related to something I removed from the test, it lets you modify objects or environments temporarily during a test. More useful for a test of a function that is doing something more complex

tests/test_station.py
```diff
@@ -66,3 +66,132 @@ def test_get_table():
...
+                            "date_to": "2020-12-31",
+                            "some_value": np.nan
+                        }
+                    ]
+                }
+            ]
+        }]
+    }
+    
+    mm = bw.MeasurementStation(test_data)
```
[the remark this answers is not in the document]
olivia-bentley: updated to use existing jsons

brightwind/load/station.py
olivia-bentley: does that mean that we want to remove the lines of code that perform .dropna which are responsible for this behaviour?

BiancaMorandi: @olivia-bentley yes please remove dropna

BiancaMorandi:
@stephenholleran merge request for this issue is ready for your review/approval. We have made changes as reported on comments on issue #421

stephenholleran (approved):

Issue #450, opened 2024-11-20
[__get_names] Returns wind speed and vertical wspd cols


rach185:
**Describe the bug**
When using the private function `__get_names`, if the `measurement_type_id = 'vertical_wind_speed'`, then all wind speed and vertical wind speed columns names are returned. 

**To Reproduce**

```
meas_sta_uuid = 'f4af1419-147e-420d-b71a-431df3bf4fe2'
meas_station = bw.MeasurementStation(bw.LoadBrightHub.get_data_model(measurement_station_uuid=meas_sta_uuid))
measurement_type_id= 'vertical_wind_speed'
meas_station.measurements._Measurements__get_names(measurement_type_id)
```

**Expected behavior**
The wind speed columns should not be returned when the `measurement_type_id` is `vertical_wind_speed`. If the  `measurement_type_id = ['vertical_wind_speed']` then the operation is as expected.

**Screenshots**
![image](https://github.com/user-attachments/assets/bf641a66-a726-4737-b871-759fc3224371)

BiancaMorandi:
I think that this function shouldn't be private as it is useful to be able to extract the measurement names by measurement type

olivia-bentley:
work on this

stephenholleran:
> [@stephenholleran](https://github.com/stephenholleran) I think that this function shouldn't be private as it is useful to be able to extract the measurement names by measurement type, what do you think?

It could be. So it would be used as outlined below?
`meas_station.measurements.get_names(measurement_type_id)`

BiancaMorandi:
> It could be. So it would be used as outlined below?
> `meas_station.measurements.get_names(measurement_type_id)`

yes, it can be used as for your suggestion, or another option is to update the `meas_station.measurements.names` function to accept measurement type as an option. The `meas_station.measurements.names` function at the moment is using the private `_get_names` function for pulling out all meas names but the measurement_type is to NOT given as an optional input. I suggest the below, @stephenholleran are you happy with this?
- update `meas_station.measurements.names` to be `meas_station.measurements.get_names`
- add option of giving measurement_type as input, if measurement_type=None then pull out all meas names as for current behaviour of `meas_station.measurements.names`

stephenholleran:
Ah sorry, I didn't think of that. Just update the `names()` to accept the measurement type. That should be easier.

BiancaMorandi:
[...]

PR #472, opened 2025-03-18
Iss450 fix vertical windspeed column issue

olivia-bentley:


brightwind/load/station.py
```diff
@@ -1071,9 +1071,10 @@ def __get_names(self, measurement_type_id=None):
         :param measurement_type_id: The measurement_type_id to filter for the names.
         :type measurement_type_id:  str or None
         :return:                    The list of names.
-        :rtype:                     list(str)
+        :rtype:                     str | list[str]
```
BiancaMorandi, line 1074: I don't get why you made this update. The output of the function before and after your fix is a list(str), it will never happen that it is a str as `names` is defined as a list on function e.g. `names = []`. Did you mean to update the input 'measurement_type_id' type?

brightwind/load/station.py
```diff
@@ -1071,9 +1071,10 @@ def __get_names(self, measurement_type_id=None):
...
         """
         names = []
+        measurement_type_id = measurement_type_id if isinstance(measurement_type_id, list) else [measurement_type_id]
```
BiancaMorandi, line 1077: probably this is not the best way to fix this bug, this fix has added an additional bug as doesn't handle cases where `measurement_type_id=None` e.g if I run `mm1.measurements._Measurements__get_names(measurement_type_id=None)` I get an empty list, but I should get a list of all measurement names instead . Can you fix this?

image

Based on what stated on docstring, it looks like this function was initially made for working with only one measurement_type (str) or all (None) and this is also the behaviour of other functions calling this (e.g `get_heights`). @stephenholleran do you want this function to be updated to work with `measurement_type_id ` both str and list(str) e.g pull out names also for multiple measurement types? 

If we keep the current behaviour accepting only `measurement_type_id `as as str then I would suggest to fix this bug editing the line highlighted below to be as below and raise an error if `measurement_type_id ` is not str or  None

`if meas_type is not None and meas_type == measurement_type_id`

image

tests/test_station.py
```diff
@@ -66,3 +66,17 @@ def test_get_table():
         if value is not '-':
             assert value in mm2.measurements.properties[4].values()
 
+
+def test_get_names():
```
BiancaMorandi, line 70: after updating `names `function as for Stephen's request then you can probably change this test to be for `names` and maybe leave few tests for the private `_get_names` function just to be safe

can you add tests also for when `measurement_type_id=None`?

tests/test_station.py
```diff
@@ -66,3 +66,17 @@ def test_get_table():
...
+    mm1 = bw.MeasurementStation(bw.demo_datasets.floating_lidar_iea43_wra_data_model_v1_2)
+    measurement_type_id1= 'vertical_wind_speed'
+    names1 = mm1.measurements._Measurements__get_names(measurement_type_id1)
+    assert names1 == ['VSpd_80']
+    
+    measurement_type_id2= 'wind_speed'
+    names2 = mm1.measurements._Measurements__get_names(measurement_type_id2)
+    assert names2 == ['Spd_80']
+    
+    measurement_type_id3= ['wind_speed', 'vertical_wind_speed']
```
BiancaMorandi, line 80: this test is needed only if we want the function to accept list(str) see question I have asked Stephen above

BiancaMorandi (changes requested):
@olivia-bentley see my comments

brightwind/load/station.py
stephenholleran: > @stephenholleran do you want this function to be updated to work with measurement_type_id  both str and list(str) e.g pull out names also for multiple measurement types?

Not the plan but might be useful. I would do that as an enhancement in another issue. I think we should just fix the bug here by using the "==" as opposed to "in" so the user would need to type out the full measurement type name.

BiancaMorandi: ok, @olivia-bentley can you make updates according to Stephen's comment?

tests/test_station.py
BiancaMorandi: we decided to keep `names` function as it is and change `get_names` to be public, so ok to have tests for get_names

BiancaMorandi:
@stephenholleran merge request for this issue is ready for your review/approval. We have made changes as reported on comments on issue #450

stephenholleran (approved):

Issue #461, opened 2025-03-11
[LoadBrightHub] Add new function to pull the cleaning rule log from BrightHub


BiancaMorandi:
**Is your feature request related to a problem? Please describe.**
The BrightHub wind resource data management platform is opening up a new API to allow users to pull the cleaning rule logs for a particular measurement station. Without this new function, a user would have to log into BrightHub and manually download the cleaning rule log file and then load that file into the brightwind library.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #478, opened 2025-03-19
iss #461 add get_cleaning_rules to LoadBrightHub

stephenholleran:
#461

### DIFF
--- a/brightwind/load/load.py
+++ b/brightwind/load/load.py
@@ -1450,7 +1450,52 @@ class LoadBrightHub:
             # Handle all request-related errors
             raise RuntimeError(f"An error occurred while fetching the cleaning log: {e}")
         return pd.read_csv(StringIO(cleaning_log_response.text))
-      
+
+    @staticmethod
+    def get_cleaning_rules(measurement_station_uuid):
+        """
+        Get the cleaning log from BrightHub for a particular measurement station.
+
+        :param measurement_station_uuid: A specific measurement station's uuid.
+        :type measurement_station_uuid:  str
+        :return:                         The cleaning logs for the measurement station.
+        :rtype:                          list(dict)
+
+        **Example usage**
+        ::
+            import brightwind as bw
+
+        To get the cleaning rules for a specific measurement station
+        ::
+            measurement_station_uuid='9344e576-6d5a-45f0-9750-2a7528ebfa14'
+            cleaning_rules_json = bw.LoadBrightHub.get_cleaning_rules(measurement_station_uuid)
+
+        Applying the cleaning to the timeseries data.
+        ::
+            data = bw.LoadBrightHub.get_data(measurement_station_uuid)
+
+            # Apply the cleaning rules to the data resulting in a dataset that is ready to work with.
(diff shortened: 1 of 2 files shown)


stephenholleran:
Hi @olivia-bentley,

Could you please review this PR to pull cleaning rules from BrightHub?

Thanks,

tests/test_load.py
```diff
@@ -112,3 +112,11 @@ def test_load_brighthub():
...
     for col in ['Spd80mN', 'Spd80mS', 'Dir78mS']:
         assert col in data_columns
+
+    # To get cleaning log
+    cleaning_log_df = bw.LoadBrightHub.get_cleaning_log(measurement_station_uuid=measurement_station_uuid)
+    assert len(cleaning_log_df) != 0
+
+    # To get cleaning rules
+    cleaning_rules_json = bw.LoadBrightHub.get_cleaning_rules(measurement_station_uuid=measurement_station_uuid)
+    assert cleaning_rules_json[0]['measurement_location_uuid'] == measurement_station_uuid
```
olivia-bentley, line 122: should we add a check against the schema to check the contents is as expected?

stephenholleran: I did think of that but seen as it is coming from BrightHub it has already been validated so I don't think there is a need to do it again.

olivia-bentley: since there are checks for valid responses then as long as it contains a uuid it should be of the right format

brightwind/load/load.py
```diff
@@ -1450,7 +1450,53 @@ def get_cleaning_log(measurement_station_uuid):
...
+        if response.status_code == 403 and 'details' in response_json:  # insufficient permissions or download limit
+            raise ValueError(f"{response_json.get('error', '')}. {response_json.get('details', '')}")
+        if response.status_code == 404 and 'details' in response_json:  # requested resource not found
+            raise ValueError(f"{response_json.get('error', '')}. {response_json.get('details', '')}")
+        if response.status_code == 500 and 'details' in response_json:  # unexpected server error
+            raise ValueError(f"{response_json.get('error', '')}. {response_json.get('details', '')}")
+        elif 'error' in response_json:
+            raise ValueError(f"Unexpected error: Status Code {response.status_code}. {response.text}")
+
+        return response_json
```
olivia-bentley, line 1498: should we check that it is a valid json and not an empty response. And move the .json() to after the status code checks in case it doesn't convert properly

stephenholleran: @olivia-bentley 
If there are no cleaning rules for a station then it is an empty JSON, so an empty JSON is valid. If it doesn't convert then there is a bigger problem that I am not sure if we should be capturing.

If we did move the `response.json()` to after the status codes we would have to run that anyway to get the error details out.

olivia-bentley: You could print response.text if it doesn't convert to aid debugging, but maybe that isn't necessary from our perspective here

stephenholleran: I think we are ok here.

Issue #462, opened 2025-03-11
[apply_cleaning_rule] Add new function to apply the cleaning rule as stated on brighthub cleaning rule log


BiancaMorandi:
**Is your feature request related to a problem? Please describe.**
The BrightHub wind resource data management platform is opening up a new API to allow users to pull the cleaning rule logs for a particular measurement station. We want to have a function for applying these rules to the data based on the downloaded rules on the json file.

**Describe the solution you'd like**
we want to have a similar function than `bw.apply_cleaning` but that applies to cleaning rules.

BiancaMorandi:
@stephenholleran merge request for this issue is ready for you to review/approve. As part of this issue we have addressed the below:

- added new function `apply_cleaning_rules` and private subfunction `_apply_cleaning_rule` 
- added `cleaning_rules_schema` file into `load` folder. There is an open question for you to reply regarding having a version on the schema. Happy to discuss.
- added `demo_cleaning_rules_file.json` for testing
- added `is_file_extension` function in `utils` for checking file extension else than if is_file
- added `check_schema` function in `utils` for validating the schema
- added tests for `apply_cleaning_rules` and `check_schema`
- updated changelog

FYI @olivia-bentley

PR #482, opened 2025-03-20
iss #462 add functionality to apply cleaning rules

olivia-bentley:


BiancaMorandi (changes requested):
@olivia-bentley I did a first review only based on code. I will have to do some random tests of function in order to finish the review.

brightwind/load/load.py
```diff
@@ -2238,6 +2240,109 @@ def apply_cleaning(data, cleaning_file_or_df, inplace=False, sensor_col_name='Se
...
+    :param condition_col: Column that the cleaning rule is based on
+    :type condition_col: str
+    :param target_cols: Column to apply the cleaning rule to clean the data of
+    :type target_cols: List[str]
+    :param threshold: Threshold value to use in the cleaning rule
+    :type threshold: float
+    :param op_code: Operator code (1-6)
+    :type op_code: int
+    :param replacement_value: Value or string to replace the data in the target_cols with
+    :type replacement_value: str | np.nan
```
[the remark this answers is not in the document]
olivia-bentley: It can accept a string or np.nan if you were to supply "NaN" to this function it would fill the values with "NaN". But if you are using it with apply_cleaning_rules this wouldn't happen. Would there be a case where you would want to use it independently and want to convert to np.nan within this function instead?

brightwind/load/load.py
```diff
@@ -2238,6 +2240,109 @@ def apply_cleaning(data, cleaning_file_or_df, inplace=False, sensor_col_name='Se
...
+    return result_df
+
+
+def apply_cleaning_rules(data, cleaning_rules_file, inplace=False, replacement_text='NaN'):
+    """
+    Apply cleaning to a timeseries DataFrame using cleaning rules either from file or from an API call to retrieve the
+    file. The flagged data will be replaced with replacement_text values which then do not appear in any plots or affect 
+    calculations.
+
+    If the 'Stop' timestamp is missing, the data from the 'Start' until the end of the timeseries will be flagged.
```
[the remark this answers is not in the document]
olivia-bentley: I had originally planned to keep that section of code but subsequently deleted it, will remove now

brightwind/load/load.py
```diff
@@ -2238,6 +2240,109 @@ def apply_cleaning(data, cleaning_file_or_df, inplace=False, sensor_col_name='Se
...
+
+    :param data:                    Data to be cleaned.
+    :type data:                     pandas.DataFrame
+    :param cleaning_rules_file:     File path of the json file or a dictionary which contains the cleaning rules to apply.
+    :type cleaning_rules_file:      str | None
```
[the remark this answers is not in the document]
olivia-bentley: now updated to str | dict | None and code updated to handle this

brightwind/load/load.py
```diff
@@ -2238,6 +2240,109 @@ def apply_cleaning(data, cleaning_file_or_df, inplace=False, sensor_col_name='Se
...
+            return TypeError("Can't recognise the cleaning_rules_file. Please make sure it is a file path or None.")
+    # else:
+        # TODO: implement the API call here
+
+    if replacement_text == 'NaN':
+        replacement_text = np.nan
+    
+    for cleaning_rule in cleaning_json:
+        columns_to_clean = [column_name['assembled_column_name'] for column_name in cleaning_rule['rule']['clean_out']]
+        condition_variable = cleaning_rule['rule']['conditions']['assembled_column_name']
```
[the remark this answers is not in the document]
olivia-bentley: Where is condition_column_name defined?

brightwind/load/load.py
```diff
@@ -2238,6 +2240,109 @@ def apply_cleaning(data, cleaning_file_or_df, inplace=False, sensor_col_name='Se
...
+    if inplace is False:
+        data = data.copy(deep=True)
+
+    if cleaning_rules_file:
+        if isinstance(cleaning_rules_file, str):
+            with open(cleaning_rules_file) as file:
+                cleaning_json = json.load(file)
+        else:
+            return TypeError("Can't recognise the cleaning_rules_file. Please make sure it is a file path or None.")
+    # else:
```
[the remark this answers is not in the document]
stephenholleran: @BiancaMorandi yes do that.
Once we have the function working we can add it as a method.
Is this question out of date as I don't see the line of code this question would relate to.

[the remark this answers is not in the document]
olivia-bentley: it was when cleaning_rules_file_or_dict was set to None I had left a TODO comment in a couple of lines below this. I've now removed this. Are there any more comments to address on this branch? @stephenholleran @BiancaMorandi

BiancaMorandi:
Should this function validate json against a schema?

brightwind/utils/utils.py
```diff
@@ -2,10 +2,15 @@
...
+import json
+from jsonschema import validate
+from jsonschema.exceptions import ValidationError
+import brightwind as bw
 
 __all__ = ['slice_data',
            'validate_coverage_threshold',
-           'is_file']
+           'is_file',
+           'check_cleaning_rule_schema']
```
BiancaMorandi, line 13: change name to be the generic one, ok to be public

stephenholleran:
Hi @olivia-bentley,

I think this file `cleaning_rule_create.schema.json` should be located where it is used i.e. under the `load` folder. I wouldn't consider it a demo file. It is a schema that is used in the code.

olivia-bentley:
OK I'll move it now, that makes sense @stephenholleran

brightwind/load/load.py
```diff
@@ -2238,6 +2249,111 @@ def apply_cleaning(data, cleaning_file_or_df, inplace=False, sensor_col_name='Se
...
+    Load data:
+        data = bw.load_csv(bw.demo_datasets.demo_data)
+        cleaning_rules_file_or_dict = bw.demo_datasets.demo_cleaning_rules_file
+
+    To apply cleaning to 'data' and store the cleaned data in 'data_cleaned':
+        data_cleaned = bw.apply_cleaning_rules(data, cleaning_rules_file_or_dict)
+        print(data_cleaned)
+
+    To modify 'data' and replace it with the cleaned data:
+        bw.apply_cleaning(data, cleaning_rules_file_or_dict, inplace=True)
```
stephenholleran, line 2314: @olivia-bentley update to apply_cleaning_rules. Missing 'rules'.

olivia-bentley: I'm confused where is rules missing from?

brightwind/load/load.py
```diff
@@ -2238,6 +2249,111 @@ def apply_cleaning(data, cleaning_file_or_df, inplace=False, sensor_col_name='Se
...
+    ::
+        import brightwind as bw
+
+    Load data:
+        data = bw.load_csv(bw.demo_datasets.demo_data)
+        cleaning_rules_file_or_dict = bw.demo_datasets.demo_cleaning_rules_file
+
+    To apply cleaning to 'data' and store the cleaned data in 'data_cleaned':
+        data_cleaned = bw.apply_cleaning_rules(data, cleaning_rules_file_or_dict)
+        print(data_cleaned)
```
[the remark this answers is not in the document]
olivia-bentley: now added

brightwind/utils/utils.py
```diff
@@ -176,4 +183,67 @@ def is_file(file_or_folder):
...
+        if extension.lower() == extension_required:
+            return True
+        else:
+            raise ValueError(f"File extension must be {extension_required}, got: {extension}")
+    else:
+        raise ValueError(f"Input must be a {extension_required} file, got: {file_or_folder}")
+
+
 
+def check_schema(json_to_check, schema):
```
BiancaMorandi, line 208: can you add tests for this function under test_utils.py?

brightwind/utils/utils.py
```diff
@@ -176,4 +183,67 @@ def is_file(file_or_folder):
...
+
+    :param cleaning_json:   The JSON data to validate
+    :type cleaning_json:    dict
+    :param schema:          The JSON schema to validate
+    :type schema:           str | dict
+    :return:                List of validation results, each containing:
+                                - item_index (int): Index of the item in the list or 0 if single item
+                                - is_valid (bool): True if validation passes, False otherwise
+                                - error_message (str): Error message if validation fails, empty string otherwise
+    :rtype:                 bool
```
BiancaMorandi, line 220: add example of usage on docstring

BiancaMorandi (commented):
@olivia-bentley see my comments

brightwind/load/load.py
```diff
@@ -2238,6 +2249,112 @@ def apply_cleaning(data, cleaning_file_or_df, inplace=False, sensor_col_name='Se
...
+        if utils.is_extension(cleaning_rules_file_or_dict, ".json"):
+            with open(cleaning_rules_file_or_dict) as file:
+                cleaning_json = json.load(file)
+    elif isinstance(cleaning_rules_file_or_dict, List):
+        if not all(isinstance(item, dict) for item in cleaning_rules_file_or_dict):
+            raise TypeError("All elements in the cleaning_rules_file_or_dict list must be dictionaries.")
+        cleaning_json = cleaning_rules_file_or_dict
+    else:
+        return TypeError("Can't recognise the cleaning_rules_file_or_dict. Please make sure it is a file path, "
+        "dictionary or None.")
```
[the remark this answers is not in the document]
BiancaMorandi: ok I did tests and it accepts only str or list(dict). I have made relevant updates.

brightwind/utils/utils.py
BiancaMorandi: actually no need for this as I see that you have added tests for this in test_load.py when checking cleaning_rules schema

BiancaMorandi: actually no need for this as I see that you have added tests for this in test_load.py when checking cleaning_rules schema

brightwind/load/load.py
```diff
@@ -26,8 +28,17 @@
            'LoadBrightHub',
            'load_cleaning_file',
            'apply_cleaning',
+           'apply_cleaning_rules',
            'apply_cleaning_windographer']
 
+OPERATOR_DICT = {
```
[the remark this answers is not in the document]
olivia-bentley: I thought it might be used elsewhere in future. Also because it contains constants I thought it should be defined outside of the scope of the function which would avoid it being recreated on each function call, but I am happy to move it

BiancaMorandi:
merge request ready for @stephenholleran review/approval. Updates made as detailed on iss #462

brightwind/utils/utils.py
```diff
@@ -176,4 +184,76 @@ def is_file(file_or_folder):
...
+    :param file_or_folder:      The file or folder path.
+    :type file_or_folder:       str
+    :param extension_required:  The file extension needed.
+    :type extension_required:   str
+    :return:                    If is a file with desired extension type.
+    :rtype:                     bool
+    """
+    if is_file(file_or_folder):    
+        _, extension = os.path.splitext(file_or_folder)
+        if extension.lower() == extension_required:
```
stephenholleran, line 200: @olivia-bentley  Should you also make the `extension_required` lower case too?

stephenholleran:
Hi @olivia-bentley,

1. Would `validate_json` be more appropriate than `check_schema`? You use the word 'validate' in the docstring as that is what the function is doing. The library it is using also uses the verb validate. The only way to validate JSON is against a JSON Schema therefore I'm thinking that we don't need to mention schema in the name either. We are also not checking the schema but checking JSON "against" a schema so could be `check_against_schema` either. What do you think?
2. This function seems to just return an error msg on the first issue found. Is there a way for it to return all the issues in one go so the user has some idea of the amount of work needed for it to pass?
3. The cleaning rule schema is called `cleaning_rule_create.schema.json`. Could we drop the word 'create' as this is to do with "creating" the rule in a BrightHub context? We can change the name especially if we have already modified it from the original source.

Still reviewing, will continue tomorrow.

brightwind/utils/utils.py
olivia-bentley: I had assumed it would be supplied in lower case but you are right it's a good idea to convert it in case

brightwind/load/load.py
```diff
@@ -2284,6 +2296,121 @@ def apply_cleaning(data, cleaning_file_or_df, inplace=False, sensor_col_name='Se
...
+        raise TypeError("Can't recognise the cleaning_rules_file_or_list. Please make sure it is a file path "
+                        "or a list(dict).")
+    
+    cleaning_json_is_valid = True
+    for cleaning_rule in cleaning_json:
+        if utils.validate_json(cleaning_rule, cleaning_rules_schema) is False:
+            cleaning_json_is_valid = False
+    if cleaning_json_is_valid is False:
+        raise ValueError("There is a problem with the validity of the supplied JSON please check the errors above")
+
```
stephenholleran, line 2398: Would this be cleaner for the above block?
```python
for cleaning_rule in cleaning_json:
   if utils.validate_json(cleaning_rule, cleaning_rules_schema) is False:
      raise ValueError("There is a problem with the validity of the supplied JSON please check the errors above.")
```

stephenholleran:
@olivia-bentley am I missing something or does this not work for you either? When I want to replace, add the cleaning directly to my dataframe, I would expect that `data` would then have the cleaning applied. But it doesn't seem to be?
image

brightwind/load/load.py
olivia-bentley: My thinking was that this would allow the user to see any errors in any of the rules, whereas this would raise an error at the first issue. If pushed a cleaner version of my method, do you think the new version is OK?

stephenholleran: Ok, makes sense we can leave it.

> If pushed a cleaner version of my method, do you think the new version is OK?

Not sure what you mean by this?

olivia-bentley: Typo sorry, I made a change to the way I wrote the original code so it reads a little more cleanly and I pushed that change

brightwind/load/load.py
```diff
@@ -2284,6 +2296,156 @@ def apply_cleaning(data, cleaning_file_or_df, inplace=False, sensor_col_name='Se
...
+    op = OPERATOR_DICT[comparison_operator_id]
+    mask = op(df[condition_col], comparator_value)
+    if date_to:
+        date_filter = (df.index >= date_from) & (df.index < date_to)
+    else:
+        date_filter = (df.index >= date_from)
+
+    for i, col in enumerate(target_cols):
+        mask_date_range = pd.Series(False, index=df.index)
+        mask_date_range = mask & date_filter
```
stephenholleran, line 2343: @olivia-bentley this first assignment of mask_date_range is overwritten by the next one. So, is there any need for this line?

stephenholleran (approved):

Issue #452, opened 2024-12-13
bw.monthly_means Legend Fix


DOhAW:
I have noticed that the bw.monthly_means plot tends to have a legend that blocks the plotlines when used for LiDARs.
![image](https://github.com/user-attachments/assets/d482743c-e299-4579-9678-06a072c4139f)

Would it be useful to change this function so that the legend appears outside the plot? See below example. 
![image](https://github.com/user-attachments/assets/c3cf8da1-f354-48df-9d85-6f3ca7b63d4b)

BiancaMorandi:
@stephenholleran merge request for this issue is ready for you to review/approve. As part of this issue we have addressed the below:

- fixed the bug for `plot_monthly_means` function (and `monthly_means `function calling it) iss #413 :
   - the input `ylabel` is now used also if `coverage` is not None - see first plot below
   - legend is now shown also for if `coverage` is not None (`return_coverage`=True) - see first plot below
 - enhancements to functions were made to accept other inputs below:
   -  `legend` to decide to show or not the legend
   - `external_legend` to decide if legend is inside or outside and above the plot area -> see 2nd and 4th plots below
   - `show_grid `to decide if showing the grid on plot (this applies only to when `return_coverage`=False as grid was already shown when `return_coverage`=True) -> see 5th plot below
   - `xtick_delta` to set the frequency of x ticks and solve issue of having unreadable x_tick labels when plot is generated for several months of data or having date inbetween monthly dates (e.g 2018-10-15, 2018-11-15) when plot is generated for few months  -> see 3rd plot below
   - added tests
   - updated changelog

![Image](https://github.com/user-attachments/assets/56288570-0763-4869-812d-48cc7f8f710b)

![Image](https://github.com/user-attachments/assets/25c412f9-621a-4525-b43d-8998b333f633)

![Image](https://github.com/user-attachments/assets/e8832410-445d-4baf-ba6e-ee78e11b160a)

![Image](https://github.com/user-attachments/assets/638e7bf4-d71a-40d9-99d8-e25242f3df27)

![Image](https://github.com/user-attachments/assets/a846f155-2292-478b-a165-ff5a7f150f90)

stephenholleran:
@BiancaMorandi why not have the `show_grid` as True by default?

BiancaMorandi:
> [@BiancaMorandi](https://github.com/BiancaMorandi) why not have the `show_grid` as True by default?
> 
[...]

PR #486, opened 2025-03-24
Iss452 monthly means legend fix

olivia-bentley:


brightwind/analyse/plot.py
```diff
@@ -308,9 +320,17 @@ def plot_monthly_means(data, coverage=None, ylbl=''):
...
             ax.patch.set_visible(False)
             ax2.set_ylabel('Coverage [-]')
+            ax.set_ylabel(ylbl)
+            if show_legend:
+                ax.legend([data.name])
             ax2.yaxis.tick_right()
             ax2.yaxis.set_label_position("right")
-            plt.close()
+            ax.set_xticks(xticks[(xticks >= start_date) & (xticks <= end_date)])
+            ax.xaxis.set_major_formatter(DateFormatter("%Y-%m-%d"))
```
BiancaMorandi, line 329: is there a reason why you have changed the format from "%b %Y" to "%Y-%m-%d"?

brightwind/analyse/plot.py
```diff
@@ -308,9 +320,17 @@ def plot_monthly_means(data, coverage=None, ylbl=''):
             ax2.set_axisbelow(True)
             ax.patch.set_visible(False)
             ax2.set_ylabel('Coverage [-]')
+            ax.set_ylabel(ylbl)
+            if show_legend:
+                ax.legend([data.name])
```
BiancaMorandi, line 325: in this case legend should include both the timeseries and the bar dataset. At the moment, the function is showing only legend for timeseries (e.g top right on example below) and I would expect to see legend as shown on bottom left on screenshot, with label as "Data coverage"
image

BiancaMorandi (dismissed):
@olivia-bentley  see my comments below

brightwind/analyse/plot.py
```diff
@@ -265,27 +265,39 @@ def _colormap_to_colorscale(cmap, n_colors):
...
 
-    ax.set_xticks(data.index)
-    ax.xaxis.set_major_formatter(DateFormatter("%b %Y"))
+    start_date = data.index[0]
+    end_date = data.index[-1]
+    xticks = pd.date_range(
+        start=pd.Timestamp(year=start_date.year, month=start_date.month, day=1), end=end_date, freq=xtick_delta
+        )
+    ax.set_xticks(xticks[(xticks >= start_date) & (xticks <= end_date)])
+    ax.xaxis.set_major_formatter(DateFormatter("%Y-%m-%d"))
```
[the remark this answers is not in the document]
olivia-bentley: This was related to the option to adjust what the ticks are. If we leave it as month year, then I find it misleading if it doesn't start at the 1st of the month or if weekly or another delta is chosen by the user @BiancaMorandi

brightwind/analyse/plot.py
olivia-bentley: see comment above

brightwind/analyse/plot.py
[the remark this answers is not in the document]
BiancaMorandi: ok, I see. Yes, I agree that it makes sense to use "%Y-%m-%d" format then

brightwind/analyse/plot.py
olivia-bentley: Do we want it as bottom left for all cases or just when coverage is on?

olivia-bentley: OK in that case all comments are now addressed

BiancaMorandi:
merge request ready for @stephenholleran review/approval. Updates made as detailed on iss #452

stephenholleran (approved):

stephenholleran (approved):

PR #487, opened 2025-03-25
Iss459 fix distribution bug

olivia-bentley:


brightwind/analyse/analyse.py
```diff
@@ -555,11 +555,22 @@ def _derive_distribution(var_to_bin, var_to_bin_against, bins=None, aggregation_
...
-    var_to_bin = var_to_bin.dropna()
-    var_to_bin_against = var_to_bin_against.dropna()
+
+    if var_to_bin.isnull().all() or (var_to_bin == np.inf).all():
+        raise ValueError(('Cannot derive distribution of {0} as is an empty pandas.Series ' +
+                          'or contains only NaN or Inf values.').format(var_to_bin.name))
+
+    if var_to_bin_against.empty or var_to_bin_against.isnull().all() or (var_to_bin == np.inf).all():
+        raise ValueError(('Cannot derive distribution with respect to {0}: {0} ' + 
+                         'is an empty pandas.Series or contains only NaN or Inf values.').format(var_to_bin_against.name))
```
stephenholleran, line 565: @olivia-bentley 
Not sure why you list the `var_to_bin_against` twice or is the first first one "respect to {0}" supposed to be `var_to_bin` so it is consistent with the previous error?

image

stephenholleran: I have made some changes already so let me know and I can make the change.

BiancaMorandi: @stephenholleran it is actually me that made that edit initially. We should have var_to_bin_against only once and on the if statement it should be (var_to_bin_against  == np.inf).all(). Do you want me to make the edit or you can fix this?

stephenholleran: I'll do it, thanks.

stephenholleran (approved):

Issue #451, opened 2024-12-04
[apply_device_orientation_offset] Add function for applying device orientation adj for remote sensing devices


BiancaMorandi:
We should generate a function that is correcting wind direction data measured by remote sensing devices based on device orientation info.

I have drafted below an example function created using as baseline the `bw.apply_wind_vane_deadband_offset` function. Note that this function works only with logger offset changes and wont give correct values if a device orientation change is also recorded. As part of this issue the function will need to be refactored to handle also device orientation changes scenarios.

```
from brightwind.load.station import DATE_INSTEAD_OF_NONE
def apply_device_orientation_offset(data, meas_station_data_model ,measurements, inplace=False):
    """
    NOTE THAT THIS FUNCTION CAN HANDLE ONLY ONE DEVICE ORIENTATION, IF MULTIPLE CHANGES ARE SET FOR DEVICE ORIENTATION THEN
    THIS FUNCTION WONT GIVE CORRECT RESULTS - BETTER FUNCTION NEEDS TO BE WRITTEN TO HANDLE DEVICE ORIENTATION CHANGES!!! 
    
    **Example usage**
    ::
    
    mm1 = bw.MeasurementStation(bw.LoadBrightHub.get_data_model(measurement_station_uuid=meas_station_uuid))
    data = bw.LoadBrightHub.get_data(measurement_station_uuid=meas_station_uuid)

    Adjust wind directions by the device orientation:
        bw.apply_device_orientation_offset(data, mm1, mm1.measurements)
        print('\nWind direction device orientation offset adjustment is completed.')
        
    """
    # Depending on what is sent, get wdir properties into a list of properties
    wdirs_properties = bw.transform.transform._get_consistent_properties_format(measurements, 'wind_direction')
    if not wdirs_properties:
        raise ValueError('No wind direction measurements found.')

    # copy the data if needed
    data = data.copy(deep=True) if inplace is False else data
    wdir_in_dataset = False
    df = pd.DataFrame(data) if type(data) == pd.Series else data

    # Apply the offset
    for wdir_prop in wdirs_properties:
        name = wdir_prop['name']
        if name in df.columns:
            wdir_in_dataset = True
            date_to = wdir_prop.get('date_to')
            if date_to is None or date_to == DATE_INSTEAD_OF_NONE:
                date_to_txt = 'the end of dataset'
            else:
                date_to_txt = date_to

            deadband = meas_station_data_model[0].get('device_orientation_deg')
[...]

PR #491, opened 2025-03-26
iss #451 adding function to apply orientation adjustment for RSDs and…

olivia-bentley:
… demo files to use for testing.

I have added demo files, possibly it would be a good idea to now change the tests using floating_lidar_iea43_wra_data_model_v1_2 to use floating_lidar_celtic_iea43_wra_data_model_v1_2?


BiancaMorandi (changes requested):
@olivia-bentley see below my comments. Happy to discuss next week if anything is not clear.

I have summarised below the key comments:
- function should take as input list of direction names user wants to apply the device orientation to, instead of full measurements properties. This to cover ZX lidar scenarios where met station can be mounted separate than lidar or floating lidars
- try to structure function in order to easily add future enhancements to handle more complex scenarios for floating_lidar where we can have multiple devices for the same station for same period (each with his own device orientation)? And we want to make sure that the correct device orientation is considered for each measurement -> probably looping through `vertical_profile_properties `and then looping through wind direction `measurements `properties???
- make sure function applies correct offset also for scenarios below:
    - logger_offset_date_from >= meas_station_data_model_from and logger_offset_date_to >meas_station_data_model_to
    - logger_offset_date_from < meas_station_data_model_from and logger_offset_date_to >meas_station_data_model_to
- I suggest having a subfunction that is only deriving and applying the offset based on target orientation e.g _apply_dir_offset_target_orientation(wdir_data, logger_offset, target_orientation, date_from, date_to, target_orientation_name)

brightwind/transform/transform.py
```diff
@@ -1400,3 +1401,138 @@ def offset_timestamps(data, offset, date_from=None, date_to=None, overwrite=Fals
...
+        date_to_tmp = meas_station_data_model_to if date_to is None or date_to == DATE_INSTEAD_OF_NONE else date_to
+        if date_from <= meas_station_data_model_to and date_to_tmp >= meas_station_data_model_from:
+            device_orientation_deg = meas_station_data_model.get('device_orientation_deg')
+            offset = device_orientation_deg
+
+            apply_offset_from = date_from if date_from > meas_station_data_model_from else meas_station_data_model_from
+            apply_offset_to = date_to_tmp if date_to_tmp <= meas_station_data_model_to else meas_station_data_model_to
+
+            # Account for a logger offset
+            additional_comment_txt = 'to account for device_orientation'
```
BiancaMorandi, line 1520: as this part where the offset is calculated is common with `apply_wind_vane_deadband_offset` would it make more sense to have it on a separate subfunction?? e.g 
`_apply_dir_offset_target_orientation(wdir_data, logger_offset, target_orientation, date_from, date_to, target_orientation_name)`
where 
- wdir_data -> (dataframe or series) with direction data
- target_orientation_name -> (str) with definition of target orientation like 'deadband' or 'device orientation' to be used for the print statements

if you decide to do so then we should update the  `apply_wind_vane_deadband_offset` function to use this subfunction on a separate issue (to create on github).

image

brightwind/transform/transform.py
```diff
@@ -1400,3 +1401,136 @@ def offset_timestamps(data, offset, date_from=None, date_to=None, overwrite=Fals
...
+        mm1 = bw.MeasurementStation(bw.LoadBrightHub.get_data_model(measurement_station_uuid=meas_station_uuid))
+        data = bw.LoadBrightHub.get_data(measurement_station_uuid=meas_station_uuid)
+
+        Adjust wind directions by the device orientation:
+            bw.apply_device_orientation_offset(data, mm1, mm1.measurements)
+            print('\nWind direction device orientation offset adjustment is completed.')
+    """
+    
+    # Depending on what is sent, get wdir properties into a list of properties
+    wdirs_properties = _get_consistent_properties_format(measurements, 'wind_direction')
```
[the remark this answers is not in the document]
olivia-bentley: _get_consistent_properties_format  allows multiple input types and allows us to select the wind directions only

brightwind/transform/transform.py
```diff
@@ -1400,3 +1401,138 @@ def offset_timestamps(data, offset, date_from=None, date_to=None, overwrite=Fals
...
+    :type meas_station_data_models:     MeasurementStation
+    :param measurements:                Measurement information extracted from a WRA Data Model using bw.MeasurementStation
+    :type measurements:                 list or dict or _Measurements
+    :param inplace:                     If 'inplace' is True, the original direction data, contained in 'data', will be
+                                        modified and replaced with the adjusted direction data. If 'inplace' is False, the
+                                        original data will not be touched and instead a new DataFrame containing the adjusted
+                                        direction data is created. To store this adjusted direction data, please ensure it is
+                                        assigned to a new variable., defaults to False
+    :type inplace:                      bool, optional
+    :raises ValueError:                 _description_
```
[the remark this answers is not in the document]
olivia-bentley: Autodocstring adds this by default if an error is raised. I should have populated the description. I can remove it if it isn't something we normally do in the library

brightwind/transform/transform.py
```diff
@@ -1400,3 +1401,138 @@ def offset_timestamps(data, offset, date_from=None, date_to=None, overwrite=Fals
...
+        meas_station_data_model_to = meas_station_data_model.get('date_to')
+        if not meas_station_data_model_to:
+            meas_station_data_model_to = df.index[-1].strftime('%Y-%m-%dT%H:%M:%S')
+        date_to_tmp = meas_station_data_model_to if date_to is None or date_to == DATE_INSTEAD_OF_NONE else date_to
+        if date_from <= meas_station_data_model_to and date_to_tmp >= meas_station_data_model_from:
+            device_orientation_deg = meas_station_data_model.get('device_orientation_deg')
+            offset = device_orientation_deg
+
+            apply_offset_from = date_from if date_from > meas_station_data_model_from else meas_station_data_model_from
+            apply_offset_to = date_to_tmp if date_to_tmp <= meas_station_data_model_to else meas_station_data_model_to
```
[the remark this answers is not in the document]
olivia-bentley: You are right about the equality inconsistency - I'll make that consistent in the code.

Regarding the scenarios, I think we do cover both cases through our looping approach:

1) When date_from >= meas_station_data_model_from and date_to > meas_station_data_model_to:
   For this measurement station model, we set apply_offset_from to be date_from and apply_offset_to to be meas_station_data_model_to. Then we apply the offset for that measurement station model. When we loop to the next model, we'll apply the offset for the section that overlaps with that model (from that model's start date until date_to if it occurs within that model, otherwise to that model's end date).

2) When date_from < meas_station_data_model_from and date_to > meas_station_data_model_to:
   We don't apply the offset for this logger to the time period before this model is valid, so apply_offset_from is taken as the start date for that model. The apply_offset_to is determined as in case 1. The remaining part of the time series is corrected by the logger preceding this one (for the time period between date_to and the end date for that logger).

Happy to discuss next week

brightwind/transform/transform.py
```diff
@@ -1400,3 +1401,138 @@ def offset_timestamps(data, offset, date_from=None, date_to=None, overwrite=Fals
...
+    if wdir_in_dataset is False:
+        print('No wind direction measurement type found in the data.\n')
+    # if a Series is sent, send back a Series
+    if isinstance(data, pd.Series):
+        df = df[df.columns[0]]
+    return df
+
+def _device_orientation_offset_for_logger_offset(meas_station_data_models, df, name, date_to, date_from, logger_offset):
+    """
+    Applies adjustments to wind vanes to account for orientation of the device and device_orientation_deg
```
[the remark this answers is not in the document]
olivia-bentley: perhaps we can discuss this

brightwind/transform/transform.py
olivia-bentley: This is a sensible idea, let's do that

brightwind/transform/transform.py
```diff
@@ -1400,3 +1401,138 @@ def offset_timestamps(data, offset, date_from=None, date_to=None, overwrite=Fals
...
+    if not wdirs_properties:
+        raise ValueError('No wind direction measurements found.')
+
+    # copy the data if needed
+    data = data.copy(deep=True) if inplace is False else data
+    wdir_in_dataset = False
+    df = pd.DataFrame(data) if isinstance(data, pd.Series) else data
+
+    # Apply the offset
+    for wdir_prop in wdirs_properties:
```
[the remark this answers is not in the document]
BiancaMorandi: maybve just raise an error if there is overlapping period for verical profile properties ?

brightwind/transform/transform.py
BiancaMorandi: FYI I have made your subfunction even more generic in order to be used also by `apply_wind_vane_deadband_offset `

brightwind/transform/transform.py
```diff
@@ -1556,18 +1556,17 @@ def apply_device_orientation_offset(data, meas_station_data_models, wdir_cols=[]
...
     by apply_device_orientation_offset function and will raise an error.
     
     :param meas_station_data_models:    A simplified object to represent the IEA WRA Task 43 data model.
     :type meas_station_data_models:     brightwind.load.station.MeasurementStation
     :param df:                          DataFrame with the time series data
     :type df:                           pd.DataFrame
-    :raises ValueError:                 If overlapping date ranges with non-zero device orientation 
-                                        values are detected
+    :raises ValueError:                 If overlapping date ranges with device orientation values are detected
     :return:                            List of processed date ranges
```
BiancaMorandi, line 1568: why this function is returning list of date ranges? isn't the goal only to return the error?

brightwind/transform/transform.py
```diff
@@ -1598,10 +1597,10 @@ def _check_vertical_profiler_properties_not_overlap(meas_station_data_models, df
             # Check if date ranges overlap
             overlap = (range1['from'] < range2['to'] and range2['from'] < range1['to'])
             
-            # If ranges overlap and at least one has non-zero offset, raise error
```
BiancaMorandi: why on line 1583 device orientation if None is set to 0? shouldn't this be left as None?

BiancaMorandi:
merge request ready for @stephenholleran review/approval. Updates made as detailed on iss #451

brightwind/transform/transform.py
```diff
@@ -1400,3 +1401,276 @@ def offset_timestamps(data, offset, date_from=None, date_to=None, overwrite=Fals
...
+    
+    Date ranges are considered as [from, to) where 'from' is inclusive and 'to' is exclusive.
+    Overlapping date ranges with device orientation values are not supported 
+    by apply_device_orientation_offset function and will raise an error.
+    
+    :param meas_station_data_models:    A simplified object to represent the IEA WRA Task 43 data model.
+    :type meas_station_data_models:     brightwind.load.station.MeasurementStation
+    :param df:                          DataFrame with the time series data
+    :type df:                           pd.DataFrame
+    :raises ValueError:                 If overlapping date ranges with device orientation values are detected
```
stephenholleran, line 1567: What if there are no overlapping periods? I would have thought you have to return something, a True if everything is ok. I have a feeling this is bad practice not returning something.

brightwind/transform/transform.py
```diff
@@ -1400,3 +1401,276 @@ def offset_timestamps(data, offset, date_from=None, date_to=None, overwrite=Fals
...
+        print('No device orientation offset applied to following requested measurement(s) as no wind direction measurement '
+        'type found in `meas_station_data_models` for these: {}.'.format(utils.bold(str(col_not_in_datamodel))))
+    # if a Series is sent, send back a Series
+    if isinstance(data, pd.Series):
+        df = df[df.columns[0]]
+        data.update(df)
+    return df
+
+
+def _check_vertical_profiler_properties_not_overlap(meas_station_data_models, df):
```
stephenholleran, line 1554: Doesn't look like a great name with the negative near the end. Could it not be a True/False question so `_check_vertical_profiler_properties_overlap()`? See below comment too.

brightwind/transform/transform.py
```diff
@@ -1400,3 +1401,282 @@ def offset_timestamps(data, offset, date_from=None, date_to=None, overwrite=Fals
...
+                                                DATE_INSTEAD_OF_NONE else meas_station_data_model_from)
+                meas_station_data_model_to = meas_station_data_model.get('date_to')
+                
+                if date_to is None or date_to == DATE_INSTEAD_OF_NONE:
+                    date_to_tmp = meas_station_data_model_to
+                else:
+                    idx_pos = df.index.get_indexer([pd.Timestamp(date_to)], method='nearest')[0]
+                    date_to_tmp = df.index[idx_pos + 1].strftime('%Y-%m-%dT%H:%M:%S') if idx_pos + 1 < len(df.index) else date_to
+
+                if i == 0 or previous_name != name:
```
stephenholleran, line 1515: PyCharm is giving this issue
image

brightwind/transform/transform.py
```diff
@@ -1400,3 +1401,282 @@ def offset_timestamps(data, offset, date_from=None, date_to=None, overwrite=Fals
...
+    if logger_offset is not None and logger_offset != 0 and target_orientation is not None:
+        offset = offset_wind_direction(float(target_orientation), offset=-float(logger_offset))
+        additional_comment_txt = additional_comment_txt + ' and logger offset'
+    if offset:
+        # Create a mask for the date range with exclusive upper bound or to end of dataframe
+        if apply_offset_to is None:
+            mask = (wdir_data.index >= pd.Timestamp(apply_offset_from))
+            
+        # Apply offset only to the masked data
+        wdir_data.loc[mask] = offset_wind_direction(wdir_data.loc[mask], float(offset))
```
stephenholleran, line 1666: image

brightwind/transform/transform.py
```diff
@@ -1400,3 +1401,282 @@ def offset_timestamps(data, offset, date_from=None, date_to=None, overwrite=Fals
...
+
+    If configuration date from is equal to previous configuration date to then date ranges are considered as 
+    [from, to) where 'from' is inclusive and 'to' is exclusive.
+
+    Overlapping periods in vertical_profiler_properties with non-zero device orientation values are not supported 
+    and will raise an error.
+
+    :param data:                        Timeseries data.
+    :type data:                         pd.DataFrame or pd.Series
+    :param meas_station_data_models:    A simplified object to represent the IEA Wind Task 43 WRA Data Model.
```
stephenholleran, line 1438: I understand that the Station object may contain more than one data model but I think this is not good practice. I think we can restrict this function to just work with the first Station found.

stephenholleran:
Hi,

This is a lot of work! Thanks.

I got this error when I sent it a met_mast data model. Maybe we could come up with a better error msg?
image

brightwind/transform/transform.py
olivia-bentley: can we discuss this, would that mean not looping through wdirs_properties?

brightwind/demo_datasets/__init__.py
```diff
@@ -17,7 +17,11 @@
            'iea43_wra_data_model_schema_v1_2',
            'floating_lidar_iea43_wra_data_model_v1_2',
            'solar_iea43_wra_data_model_v1_3',
-           'sodar_iea43_wra_data_model_v1_3']
+           'sodar_iea43_wra_data_model_v1_3',
+           'demo_floating_lidar_data',
+           'floating_lidar_demo_iea43_wra_data_model_v1_3'
```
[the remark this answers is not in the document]
olivia-bentley: Can we discuss the best way to do this.

brightwind/transform/transform.py
BiancaMorandi: no this means not looping through meas_station_data_models. We can discuss

brightwind/transform/transform.py
stephenholleran: It is because it is declared in an if statement. When things are declared in an if statement it means that there can be a possibility that it doesn't get declared and that is what the warning is about.

brightwind/demo_datasets/__init__.py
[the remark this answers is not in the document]
stephenholleran: Can we reduce the size of the demo_floating_lidar one to at least a quarter of it's size and we can come back another time to clean these up.

stephenholleran (approved):

Issue #378, opened 2023-02-10
[LoadBrighthub] update plants and stations APIs with filters


stephenholleran:
Update the following LoadBrighthub functions:

- **.get_measurement_stations()** with a filter for 
   - 'measurement_station_type' e.g. get all measurement stations that are 'lidar'.
   - get measurement stations for a station name that contains a certain string.

- **.get_measurement_stations()** to return the `plant_name` instead of `plant_uuid`.

stephenholleran:
@olivia-bentley if possible could we include the filter for `measurement_station_type` for the `.get_measurement_stations()` only. We will come back to filtering for names as that might be better done in the BrightHub API and let's forget about plants.

PR #499, opened 2025-04-07
iss #378 additional functionality for get_measurement_stations to all…

olivia-bentley:
…ow filtering for device type


stephenholleran:
Hi @olivia-bentley,

Only thing here is maybe we could give a better error msg if after the filtering there was nothing found. I spelt 'lidar' wrong but the error comes from creating the dataframe.

image

Apart from that it works well. The list of types is a good addition.

stephenholleran (approved):

Issue #492, opened 2025-03-28
plot_timeseries() - allow it to plot more than 12 lines


stephenholleran:
**Is your feature request related to a problem? Please describe.**
The function `bw.plot_timeseries()` fails if there are more than 12 lines to plot. This can happen easily these days with lidars, sodars and adcps measuring more levels.

![Image](https://github.com/user-attachments/assets/993e2928-ece0-46e5-9acd-307d2ab50776)

**Describe the solution you'd like**
Allow it to plot more than 12 lines.

**Describe alternatives you've considered**
Could loop through the existing 12 colors or use a graduated color map. The graduated color map might be better so a user can see a change from top to bottom or whatever y-axis metric is used.

olivia-bentley:
@BiancaMorandi I've updated the code now with one test remaining to write. I've added in a user defined colourmap option and a brightwind colour one using the code above. How do you do the colourmap with 3 colours that you mention in your comment [COLOR_PALETTE.tertiary, COLOR_PALETTE.primary,COLOR_PALETTE.fifth]? Are you adding togther 3 separate calls to _colormap_to_colorscale

BiancaMorandi:
> [@BiancaMorandi](https://github.com/BiancaMorandi) I've updated the code now with one test remaining to write. I've added in a user defined colourmap option and a brightwind colour one using the code above. How do you do the colourmap with 3 colours that you mention in your comment [COLOR_PALETTE.tertiary, COLOR_PALETTE.primary,COLOR_PALETTE.fifth]? Are you adding togther 3 separate calls to _colormap_to_colorscale

@olivia-bentley I don't think that we need to have an additional `color_map` input. Cannot the `line_colors` do the job already? if the length of the lines to plot is bigger than the predefined list (12 colors) then use another list generated from the default color map. We should probably define a different `color_map_colors` variable (e.g `color_map_colors_graduated`) at the beginning of plot.py where we set the default red-yellow colours?

```
COLOR_PALETTE = bw.analyse.plot.COLOR_PALETTE
COLOR_PALETTE.color_map_colors_graduated= [COLOR_PALETTE.tertiary,
                                      COLOR_PALETTE.primary,COLOR_PALETTE.fifth]
color_list = bw.analyse.plot._colormap_to_colorscale(COLOR_PALETTE.color_map_colors_graduated, 
                                                         len(time_series_plot_instruments))
```
lets discuss

olivia-bentley:
[...]

PR #500, opened 2025-04-08
Iss492 plot more timeseries

olivia-bentley:


olivia-bentley:
@stephenholleran the colourmap is now reversed which plots like the below. Are you happy with how it looks?
image

stephenholleran:
@olivia-bentley I think the top one is better looking with the darker lines at the top which relate to the higher heights.

brightwind/analyse/plot.py
```diff
@@ -117,19 +117,13 @@ def __init__(self):
 
             bw.analyse.plot.COLOR_PALETTE.primary_10 = '#0a1429'
 
-            # The colors used for defining the color_map can also be reset by using the example code below:
+            # The colors used for defining any of the three defined colour maps can also be reset by using the 
+            # example code below:
```
BiancaMorandi, line 121: @olivia-bentley the method you have used is not giving what expected. The reason why we had separate color map variables for each of the 3 color maps was to be able to set these separately as for examples that were given. E.g for an assessment where I want to generate plots using the different plots using the different color maps I want to be able to set my own colors and list of colors.

BiancaMorandi: I have fixed this and other bugs

BiancaMorandi:
merge request ready for @stephenholleran review/approval. Updates made as detailed on iss #492

stephenholleran (approved):

Issue #287, opened 2021-02-24
[LoadBrightHub] optionally return the raw download from the API and not put it into a df


stephenholleran:
This would apply to the `get_plants()` and `get_measurement_stations()`.

The idea is to be able to make it easier to work with these instead of the dataframe.

stephenholleran:
@olivia-bentley if possible, could we include returning the JSON for the `get_measurement_stations()` function only?

PR #501, opened 2025-04-10
iss #287 returning a list of dictionaries as an option from get_measu…

olivia-bentley:
…rement_stations


stephenholleran:
@olivia-bentley I had to resolve a good few conflicts. Could you please check if the tests still run for you or if you notice anything I did incorrectly?

Issue #407, opened 2023-07-10
[load_csv] 'DataFrame' object has no attribute 'append'. Pandas 2.0+ removed "df.append()" function.


Ilirmc:
**Describe the bug**
The bw.load_csv function fails to load csvs from a folder, as the pandas "df.append()" function has been removed in pandas 2.0+. This should be replaced with concat.

**To Reproduce**
df = bw.load_csv(folder_path_containing_csvs)
# AttributeError: 'DataFrame' object has no attribute 'append'

**Expected behavior**
bw.load_csv() should load all the csv files from the folder and return them as a single DataFrame

stephenholleran:
Thanks @Ilirmc for highlighting. Do you want to suggest a solution or we can take a look at it ourselves?

Cheers,

cc @BiancaMorandi

stephenholleran:
@Ilirmc are you able to tackle this in the next week or so?
Thanks.

PR #517, opened 2025-07-18
iss255 tackle deprecation warnings

rm-mol:
Linked to #407 

I have been tackling deprecation warnings on my fork, and now pushing it to dev


brightwind/transform/transform.py
```diff
@@ -198,7 +196,7 @@ def _round_timestamp_down_to_averaging_prd(timestamp, period):
                                              hour=timestamp.hour)
     elif period[-1] == 'M' or period[-2:] == 'MS':
         return '{year}-{month}'.format(year=timestamp.year, month=timestamp.month)
-    elif period[-2:] == 'AS' or period[-1:] == 'A':
+    elif period[-2:] == 'YS' or period[-1:] == 'A':
```
stephenholleran, line 199: Should the 'A' be 'Y' too?

If users already use 'AS' or 'A' then wouldn't this be a breaking change? Could we include 'YS' or 'Y' as well as keeping the 'AS' and 'A'?

brightwind/analyse/analyse.py
```diff
@@ -147,10 +147,10 @@ def custom_agg(x):
                      axis=1).dropna()
 
     if aggregation_method == '%frequency':
-        counts = data.groupby([y_series.name, x_series.name]).count().unstack(level=-1)
+        counts = data.groupby([y_series.name, x_series.name],observed=False).count().unstack(level=-1)
```
stephenholleran, line 150: It would be great if we could keep consistent convention and have a space character after a comma?

stephenholleran:
Hi @r-molins-mrp,

Again, thanks a mil for working on this. I just had some small comments on the replacing 'AS' with 'YS' but thinking about it a bit more even the 'H' to 'h' might be breaking changes. I'll need to look into a bit more.

@BiancaMorandi is on annual leave at the moment but when she is back I'm sure she will take a look too.

We are going to be putting more effort into this library over the next few months. We have to figure out what releases we should be planning and what goes into them.

Thanks,

brightwind/transform/transform.py
```diff
@@ -61,17 +59,17 @@ def _freq_str_to_dateoffset(period):
     elif period[-1:] == 'S':
         as_dateoffset = pd.DateOffset(seconds=float(period[:-1]))
     else:
-        raise ValueError('"{}" period not recognized. Only units "M", "MS", "A", "AS", "W", "D", "H", "T", "min", "S" '
+        raise ValueError('"{}" period not recognized. Only units "M", "MS", "YS", "W", "D", "h", "T", "min", "S" '
```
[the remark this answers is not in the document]
r-molins-mrp: 'A' and 'AS' were removed as they have deprecations warnings and will be deprecated in `pandas` versions 3.0, see:
https://github.com/pandas-dev/pandas/pull/55479

But we could also keep 'A' and 'AS' to be safe, if you prefer it to be backward compatible.

'Y' doesn't seem to be in list of pandas frequency string anymore, which is why I didn't add it:
https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#dateoffset-objects

Tell me what you prefer then.

brightwind/analyse/analyse.py
r-molins-mrp: Of course, edited.

r-molins-mrp:
Thanks for your comments @stephenholleran, see my answers above on the A/AS/Y/YS topic.

Let me go what route you prefer and I will edit the code; I guess the safest would be to add all former cases (?)

stephenholleran:
Thanks @rm-mol again.
I made some comments there. Let me know if anything isn't clear.

rm-mol:
Thanks for your comments; I understand that maintaining back-compatibility is key here. I simply pushed from our fork the deprecation changes, but on our fork we don't focus on back-compatibility (as we just have one environment that we update regularly), therefore my proposed changes would need to be updated to reflect that. I will close the present PR for the time being, until I have a better solution for these deprecation changes then.

stephenholleran:
Thanks. Do you mean this branch though
https://github.com/mainstreamrp-dev/brightwind/tree/solve-conflicts-dev ?

brightwind/transform/transform.py
```diff
@@ -252,7 +275,7 @@ def _round_timestamp_down_to_averaging_prd(timestamp, period):
                                                                 day=timestamp.day, hour=timestamp.hour,
                                                                 minute=_round_down_to_multiple(timestamp.minute,
                                                                                                int(period[:-3])))
-    elif period[-1] == 'h':
+    elif period[-1] == 'h' or period[-1] == 'H':
```
stephenholleran: Hi @olivia-bentley,

This edit triggered me to look up the chain as to what calls this function as there is no depreciation warning. It goes up to `merge_datasets_by_period()`.

When that is called, the `_freq_str_to_dateoffset()`, where we handle the backward compatibility, isn't called in the chain down to this `_round_timestamp_down_to_averaging_prd()`. Therefore, I'm not sure that the `merge_datasets_by_period()` handles backward compatibility and would break. Later in the function it calls `average_data_by_period()` which does hit `_freq_str_to_dateoffset()` but it would probably fail before it gets there.

Could you investigate if this is the case?

Thanks,

brightwind/transform/transform.py
```diff
@@ -34,6 +34,28 @@ def _compute_wind_vector(wspd, wdir):
...
+    :type func:             Callable
+    :return:                Transformed DataFrame
+    :rtype:                 pd.DataFrame
+    """
+    if hasattr(df, 'map'):
+        # pandas >= 2.1
+        return df.map(func, **kwargs)
+    else:
+        # pandas < 2.1
+        return df.applymap(func, **kwargs)
```
stephenholleran: @olivia-bentley could we add a depreciation warning here like I did for the freq strings to let users know that they need to update?

brightwind/transform/transform.py
stephenholleran: I think searching for the word "period" in the public functions and seeing if the period argument passes through `_freq_str_to_dateoffset()` would be the thing to do to ensure we haven't missed any.

stephenholleran (approved):

PR #536, opened 2025-09-15
[Iss530] Add functions to extrapolate temperature and air density (SR14BM9)

sararafter:
Added two functions `extrapolate_temperature` and `extrapolate_density` - both using common private function `bw.utils.utils.vertically_extrapolate_at_constant_rate`. 

These functions are for extrapolating reference temperature and density values (respectively) from a `reference_height` to a `target_height` using a constant `lapse_rate`. The `lapse_rate` argument has a default value in each case.

Closes: #534 
Closes: #530

Issue #535, opened 2025-09-12
[calc_air_density] Update calc_air_density to IEC formula accounting for humidity (SR10BM7)


sararafter:
**Is your feature request related to a problem? Please describe.**
Update `calc_air_density` function to incorporate humidity measurements as suggested in IEC standards (61400-12-1).

**Describe the solution you'd like**
Update air density calculation to the following calculation:

<img width="977" height="487" alt="Image" src="https://github.com/user-attachments/assets/076f403d-4e70-4ee0-a4c1-5e098d206ca7" />

Take input temperature, pressure and relative humidity (as float or pandas.Series) and output calculated air density (as float or pandas.Series). Set default humidity = 0 to avoid breaking change for existing function.

sararafter:
To avoid breaking change, if `relative_humidity_percent` argument is None, then the `specific gas constant` argument in the function is used. The default value of `specific gas constant` is 286.9 as was previously the case, and the humidity is taken as 0 in the calculation of air density. Therefore, if `relative_humidity_percent` argument is None, the method is identical to the behaviour of this function prior to the update to allow humidity as an argument.

If `relative_humidity_percent` argument is not None, then the `specific gas constant` argument is ignored and the correct gas constants for dry air and water vapour are used. 

A deprecation warning is added to highlight that `specific gas constant` argument should be removed in the next breaking change release, in favour of using the correct gas constants for dry air and water vapour.

A deprecation warning highlights that the option to scale air density to the site elevation within `bw.calc_air_density` will be removed in the next breaking change release, in favour of calling `bw.scale_air_density_to_height` separately.

Issue #530, opened 2025-09-03
Add function to lapse temperature measurements from measurement height to a different height


sararafter:
**Is your feature request related to a problem? Please describe.**
Air density calculation should be performed with temperature and pressure measurements at the same measurement height, however the pressure and temperature sensors may not be reported at the same height. A function is needed to lapse temperature measurements to obtain temperature values at a different height (in particular, at the height at which the pressure is measured for an air density calculation, though this function should be more general).

**Describe the solution you'd like**
Take input of temperature value at a particular measurement height. Input the output_height you want to output a temperature value for. Return temperature value at  the output_height.

There are several lapse rates available for temperature. The actual lapse rate would vary with time and location (and with temperature/pressure/humidity) but for this use it might be most practical to assume a constant lapse rate?

The main options would be:
1. dry adiabatic lapse rate for of 9.8 deg C / km (which assumes no heat/energy exchange with surroundings so is unrealistic) or
2. a lapse rate of 6.5 deg C / km which is what older versions of WASP used
(Later versions of WAsP incorporate local temperature lapse rates, often derived from reanalysis data like [ERA5])

If using a constant lapse rate, option 2 is probably the most realistic, but the function could have a lapse_rate argument (with default value) if that would be useful.

**Additional context**
Do we also need a function to lapse pressure measurements to a different height?

BiancaMorandi:
@stephenholleran I have reviewed PR #536 (addressing this issue and #534 ) and I am happy for you to have a last look and merge. As part of the PR the following enhancements were made:
- added `scale_air_density_to_height` and tests
- added `scale_air_temperature_to_height` and tests
- the two functions above are using a common function `linear_transform` within `utils` for applying the scale. This is a new function and tests were added also for this.
- there is only an open question for you on the PR relative to referencing the air density lapse rate as -0.113

FYI @sararafter

stephenholleran:
Merged.

PR #538, opened 2025-09-16
[Iss535] updated calc_air_density to include humidity and follow IEC-…

sararafter:
Note this depends on `extrapolate_density` function (depends on PR #536).

I have made change to include humidity in `calc_air_density` (default humidity = 0 to avoid breaking change) as in IEC:61400-12-1 below:
image

In theory, humidity = 0 should result in the same value as the previous version of this function, with 
$$\rho = \frac{P}{TR}$$
where R is the specific gas constant.

However, the old version has default specific_gas_constant=286.9. This value is for air with some moisture in it. 

The IEC equation uses specific_gas_constant=287.05. This value is for **dry air** (with no moisture). It doesn't make physical sense to use the old default value in the new calculation as the function is clearly split into a 'dry' component (yellow) (which is what remains when humidity = 0) and the component which is multiplied by humidity (blue), which depends on the specific gas constant for water vapour (R_w).

I don't see any way to avoid a breaking change unless we keep the old value of the specific_gas_constant which is clearly incorrect in this context.
- Options could be to change the value of the specific_gas_constant argument to 287.05 by default (however this is a breaking change).
- Could add another argument like specific_gas_constant_dry_air? And use the old specific_gas_constant only when humidity=0 (this would prevent a breaking change, but in doing so would be using a gas constant based on air with moisture in it which contradicts the assumption that humidity is 0).
- Honestly I don't see any reason for the specific_gas_constant to be an argument to the function at all? Both the specific gas constants for dry air and water vapour could just be defined within the function and this argument could be removed altogether (also a breaking change).

I have also changed the TypeError messages, added units to the argument names, and I think it would make more sense for the lapse rate to be -0.000113 because the heights are input in meters, so it would be more intuitive for the lapse rate to be in kg/m3/m rather than kg/m3/km.

I don't think avoiding a breaking change should come at the expense of making physical sense, so open to ideas of how best to handle this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Additional labels for pre-release and build metadata are available as extensions
 
 ## [2.4.0-dev]
 1. Fixed pandas<3.0.0 and numpy<2.3.1 dependencies, ([#458](https://github.com/brightwind-dev/brightwind/issues/458))
+2. Updated `calc_air_density` to include relative humidity as suggested in IEC Standards (61400-12-1) ([[#535](https://github.com/brightwind-dev/brightwind/issues/535)])
 
 ## [2.3.0]
 This update brings a comprehensive set of **bug fixes** and **enhancements** across the Brightwind library. 
--- a/brightwind/analyse/analyse.py
+++ b/brightwind/analyse/analyse.py
@@ -2013,55 +2013,78 @@ def sector_ratio(wspd_1, wspd_2, wdir, sectors=72, min_wspd=3, direction_bin_arr
     return fig
 
 
-def calc_air_density(temperature, pressure, elevation_ref=None, elevation_site=None, lapse_rate=-0.113,
-                     specific_gas_constant=286.9):
+def calc_air_density(temperature_degC, pressure_hPa_mbar, relative_humidity_percent = 0, elevation_ref_m=None, elevation_site_m=None, lapse_rate=-0.113,
+                     specific_gas_constant=287.05):
     """
-    Calculates air density for a given temperature and pressure and extrapolates that to the site if both reference
-    and site elevations are given.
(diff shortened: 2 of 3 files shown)


stephenholleran:
What is the difference in the resulting air density for a typical situation if we use the correct dry air constant verses the none dry air constant?

dancasey-ie (changes requested):

BiancaMorandi:
@sararafter I have made several changes in order to revert all the breaking changes below that you have introduced for `calc_air_density`  -> see details on https://github.com/brightwind-dev/brightwind/pull/538/commits/200683dab3040d1e88d2a60257ef31cc7d4a510d:
- input argument names must be the same than previous function (e.g. we need to keep `temperature`, `pressure` etc.). These names are not reporting correct meas type name or have unit as for our latest naming convention but we need to accept this in order to avoid breaking changes
- input argument order must stay the same, this means that the additional `rel_humidity_percent` argument added must be at the end
- argument type for pre-existing variables must be the same (you can add types but not remove)
- old function was not rounding output air density when not scaled but you had add a round(5). 
- old function was rounding output air density with round(3) when scaled but you had add a round(5). I don't get why old function was applying rounding but lets keep this like this for consistency. We are anyway reporting that the scaling functionality will be deprecated and advice the user to use `scale_air_density_to_height` instead (that is not applying any rounding).

In addition to the above, I have reworded a bit docstring and warnings (e.g. you had deprecation warnings together but these should be triggered by different if statements as for my fix), added tests, fixed few tests and grouped by test type, fixed a bug when temperature or pressure or rel_humidity_percent is DataFrame. 

Please have a look at my changes, happy to discuss.

On this issue you had also merged #530, can you repeat the merge and solve conflicts after making the fixes Stephen has asked for #530.

BiancaMorandi:
@stephenholleran this PR for the #535 issue is ready for your to review.

BiancaMorandi (approved):

stephenholleran (approved):

PR #542, opened 2025-10-03
[Iss541] update scale wind spd name

sararafter:
I updated `scale_wind_speed` function name and docstring - the new name is apple_scale_factor.


brightwind/analyse/analyse.py
```diff
@@ -1324,7 +1324,7 @@ def freq_table(var_series, direction_series, var_bin_array=np.arange(-0.5, 41, 1
         else:
             freq_tab_mean = _calc_mean_speed_of_freq_tab(result)
             scale_factor = target_freq_table_mean / freq_tab_mean
-            var_series_scaled = scale_wind_speed(var_series_scaled, scale_factor)
+            var_series_scaled = apply_scale_factor(var_series_scaled, scale_factor)
```
dancasey-ie: tf.apply_scale_factor

brightwind/utils/utils.py
```diff
@@ -252,3 +254,39 @@ def validate_json(json_to_check, schema):
             print(f"Failed schema part: {error.get('schema_path')}\n")
     
     return data_is_valid
+
+
+def apply_scale_factor(data : Union[float, pd.Series, np.array],
```
BiancaMorandi: should this accept also dataframe? the scale_wind_speed function accepts also this type so the generic function called should do the same...please update here and on docstring

brightwind/utils/utils.py
```diff
@@ -252,3 +254,39 @@ def validate_json(json_to_check, schema):
...
+    # scale np.array by scale_factor of 0.5
+    bw.apply_scale_factor(np.array([0, 1, 2]), 0.5)
+    # [0, 0.5, 1]
+    
+    # scale pd.Series by scale_factor of -10
+    bw.apply_scale_factor(pd.Series([10, 20, 30, 40]), -10)
+    # [-100, -200, -300, -400]
+    """
+
+    if not isinstance(data, (float, int, pd.Series, np.ndarray)):
```
BiancaMorandi: if accepts int then you should add this type also to function and docstring type definitions and error sentence

brightwind/utils/utils.py
```diff
@@ -339,3 +340,61 @@ def linear_transform(x_target: Union[float, int, np.ndarray, pd.Series],
...
+    # 0	1.0	6	a
+    # 1	2.4	8	b
+    """
+
+    if not isinstance(data, (float, int, pd.DataFrame, pd.Series, np.ndarray)):
+        raise ValueError('data should be a float or int or pd.DataFrame or pd.Series or np.ndarray')
+    if not isinstance(scale_factor, (float, int)):
+        raise ValueError('scale_factor should be a float or int')
+    
+    if isinstance(data, pd.DataFrame):
```
BiancaMorandi: @sararafter I don't understand why we need this exception. If we want to have this exception then shouldn't this apply also to pd.Series and np.array?

BiancaMorandi (approved):

stephenholleran (approved):

Issue #545, opened 2025-10-31
[time_continuity_gaps] Add minimum time gap to show as argument to the function


olivia-bentley:
Often we post process the table generated in order to remove small time gaps depending on the records at the site. We propose to add time gap as an optional argument to the function so that the records returned can be pre-filtered.

PR #546, opened 2025-11-03
iss #545 added argument minimum_gap_length to time_continuity_gaps

olivia-bentley:
Argument minimum_gap_length added to `time_continuity_gaps` function. This facilitates filtering the returned DataFrame by a given TimeDelta.


brightwind/analyse/analyse.py
```diff
@@ -1425,7 +1431,12 @@ def time_continuity_gaps(data):
 
     continuity = pd.DataFrame({'Date From': indexes.values.flatten()[:-1],
                                'Date To': indexes.values.flatten()[1:]})
-    continuity['Days Lost'] = (continuity['Date To'] - continuity['Date From']) / pd.Timedelta('1 days')
+    continuity['Days Lost'] = (continuity['Date To'] - continuity['Date From'])
```
stephenholleran: I can see why you removed the pd.Timedelta from this line so you can filter out gaps smaller than a certain length. I think the variable should be renamed though so as to match what it represents i.e. it doesn't represent 'Days Lost' anymore as that is calculated below.

PR #549, opened 2025-11-11
[Iss539] Add alternative air density method (SR5)

sararafter:
Added four new functions which use WindPRO method / calculating air density from dew point temperature.

For more detailed references see the following:
https://help.emd.dk/knowledgebase/content/ReferenceManual/DensityOfAir.pdf
https://wahiduddin.net/calc/density_altitude.htm

1. `calc_air_density_from_vapour_prs` is for calculating air density using the ideal gas law from air pressure and vapour pressure due to the water vapour component of the air.
2. `calc_vapour_pressure_from_dewpoint` - to calculate the vapour pressure needed in `calc_air_density_from_vapour_prs` function above, one method is to use the dew point temperature.
3. `calc_vapour_pressure_from_relative_humidity`  - to calculate the vapour pressure needed in `calc_air_density_from_vapour_prs` function above, alternative method uses relative humidity. 
4. Both `calc_vapour_pressure_from_dewpoint` and `calc_vapour_pressure_from_relative_humidity` functions to calculate vapour pressure rely on a common function `calc_saturation_vapour_pressure_of_water_vapour` which calculates the saturation vapour pressure of water vapour at a given temperature using the Herman Wobus polynomial approximation.

### DIFF
--- a/brightwind/analyse/analyse.py
+++ b/brightwind/analyse/analyse.py
@@ -2135,10 +2135,14 @@ def calc_air_density(temperature: Union[float, pd.Series],
     :type rel_humidity_percent:         float or pandas.Series
     :param dew_point_temperature_degC:  Dew point temperature values in degrees Celsius. Default is None.
     :type dew_point_temperature_degC:   float or pandas.Series
-    :param calc_method:                 Method to calculate water vapour pressure. Options are:
+    :param calc_method:                 Method to calculate water vapour pressure using
+                                        bw.analyse.analyse._calc_water_vapour_pressure().
+                                        Options are:
                                         - 'IEC' (default):
-                                        Requires air temperature (temperature) and optional
+                                        Inputs are air temperature (temperature) and
                                         relative humidity (rel_humidity_percent).
+                                        If rel_humidity_percent input to bw.calc_air_density() is None,
+                                        a value of 0% is assumed for the water vapour pressure calculation.
                                         - 'HermanWobus_from_rel_humidity':
                                         Requires air temperature (temperature) and
                                         relative humidity (rel_humidity_percent).


brightwind/analyse/analyse.py
```diff
@@ -29,10 +29,17 @@
...
-           'calc_air_density']
+           'calc_air_density',
+           'calc_air_density_from_vapour_prs',
+           'calc_saturation_vapour_pressure_of_water_vapour',
+           'calc_vapour_pressure_from_dewpoint',
+           'calc_vapour_pressure_from_relative_humidity']
 
 #  Specific gas constant for dry air (J/K/kg or m2/K/s2) from ISO:2533-1975 Standard Atmosphere
-GAS_CONST_DRY_AIR = 287.05 
+GAS_CONST_DRY_AIR = 287.05
```
BiancaMorandi: this is the same constant reported on scale.py (same comment for AIR_DENSITY_LAPSE_RATE) , we should have it only in one place. Can you add a constant.py file maybe inside utils folder, move all scale pressure, scale temperature, scale/derived air density constants there (e.g. GAS_CONST_DRY_AIR, GAS_CONST_WATER, AIR_DENSITY_LAPSE_RATE etc..) and update relevant .pys to get constants from there?

brightwind/analyse/analyse.py
```diff
@@ -2245,3 +2252,295 @@ def calc_air_density(temperature: Union[float, pd.Series, pd.DataFrame],
     return air_density
```
BiancaMorandi: everywhere, can you please use air pressure, air temperature wording instead of pressure and temperature? this to use IEA task-43 definitions.

brightwind/analyse/analyse.py
```diff
@@ -2245,3 +2252,295 @@ def calc_air_density(temperature: Union[float, pd.Series, pd.DataFrame],
...
+    P_v_Pa = 100*vapour_pressure_hPa
+    P_Pa = 100*air_pressure_hPa
+    T_K = air_temperature_degC + 273.15
+    R_d = GAS_CONST_DRY_AIR
+    R_v = GAS_CONST_WATER
+    P_d_Pa = P_Pa - P_v_Pa
+    air_density = P_d_Pa/(R_d*T_K) + P_v_Pa/(R_v*T_K)
+    return air_density
+
+
```
BiancaMorandi: I would suggest to call function `calc_saturation_water_vapour_pressure `and reword a bit wording on docstring to follow this. 

I don't think that we should say that it used by windpro but if we want to say this then should say also on the main `calc_air_density_XXX`

brightwind/analyse/analyse.py
```diff
@@ -2245,3 +2252,295 @@ def calc_air_density(temperature: Union[float, pd.Series, pd.DataFrame],
...
+    air_density.tail(5)
+    # Timestamp
+    # 2017-11-23 10:10:00    1.195418
+    # 2017-11-23 10:20:00    1.196145
+    # 2017-11-23 10:30:00    1.196055
+    # 2017-11-23 10:40:00    1.195684
+    # 2017-11-23 10:50:00    1.197327
+    # dtype: float64
+    """
+    # Check dimensions of temperature, pressure and rel_humidity_percent if not float or int
```
BiancaMorandi: probably we need to have a check for  input type as well...it would be nice to have private functions doing the length and type check as there are several functions having very similar code. I leave @stephenholleran decide if he wants this...I guess that can be a future enhancement no needed now

brightwind/analyse/analyse.py
```diff
@@ -2245,3 +2252,295 @@ def calc_air_density(temperature: Union[float, pd.Series, pd.DataFrame],
...
+    # dtype: float64
+    """
+    if isinstance(relative_humidity_percent, pd.Series) or isinstance(air_temperature_degC, pd.Series):
+        if len(relative_humidity_percent) != len(air_temperature_degC):
+            raise ValueError("relative_humidity_percent and air_temperature_degC must have the same dimensions.") 
+    P_v_hPa = (relative_humidity_percent/100)*calc_saturation_vapour_pressure_of_water_vapour(air_temperature_degC)
+    return P_v_hPa
+
+
+def calc_vapour_pressure_from_dewpoint(dew_point_temperature_degC):
```
BiancaMorandi: this is exactly same function as `calc_saturation_vapour_pressure_of_water_vapour`, if we want to implement method I was suggesting for calc_air_density and have all other functions private then I don't think that we need this function

tests/test_scale.py
```diff
@@ -45,7 +45,6 @@ def test_scale_air_pressure_to_height():
             ref_height_m=2,
```
BiancaMorandi: can you undo any empty line change made for this test_scale.py? we should have two empty lines between functions

brightwind/analyse/analyse.py
```diff
@@ -2245,3 +2252,295 @@ def calc_air_density(temperature: Union[float, pd.Series, pd.DataFrame],
     return air_density
```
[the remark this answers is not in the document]
stephenholleran: Thanks Bianca.
I agree there should only be one function to calculate the air density and within that function the different method can be specified (not a Boolean as there may be more methods later). Makes it a lot easier for a wind analyst to be able to calculate air density then.

@sararafter what extra inputs would the `calculate_air_density()` need to take to do this?

brightwind/analyse/analyse.py
```diff
@@ -2206,22 +2270,42 @@ def calc_air_density(temperature: Union[float, pd.Series],
...
-                            (1 / gas_constant_air) - (1 / specific_gas_constant_water))
-    
-    # Combine to get air density
-    air_density = (1 / temp_K) * (dry_air_term - water_vapour_term)
-    
+    # IEC uses a slightly different value for specific gas constant of water vapour (461.495 vs 461.5) - handle:
+    if calc_method == 'IEC':
+        specific_gas_constant_water = 461.5
+    else:
+        specific_gas_constant_water = GAS_CONST_WATER
```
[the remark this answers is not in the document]
stephenholleran: If this doesn't make a material difference to the results/tests then I think we are fine to change it without incrementing the major version number. The difference between the two numbers is very small I expect it doesn't have an impact.

brightwind/analyse/analyse.py
```diff
@@ -2206,22 +2274,38 @@ def calc_air_density(temperature: Union[float, pd.Series],
...
-    water_vapour_term = rel_hum_decimal * vapour_press * (
-                            (1 / gas_constant_air) - (1 / specific_gas_constant_water))
-    
-    # Combine to get air density
-    air_density = (1 / temp_K) * (dry_air_term - water_vapour_term)
-    
+    # If calc_method requires dew point temperature, ensure it is always <= temperature:
+    if calc_method == 'HermanWobus_from_dew_point':
+        if dew_point_temperature_degC is None:
+            raise ValueError("dew_point_temperature_degC must be provided when calc_method is "
```
BiancaMorandi: this is a duplicate check respect what done on `_calc_water_vapour_pressure_Pa`, is this duplication needed?

Issue #569, opened 2026-01-08
[apply_wind_vane_deadband_offset] Table is not showing any values when deadband or/and logger offset value is none and other bug


BiancaMorandi:
**Describe the bug**
When the input measurements data model has None values for vane deadband orientation or/and logger offset, for a measurement, the function is not applying any adjustment and the returned table is not showing any values. 

- This is a bug as table should report values for Height [m], Vane Dead Band Orientation [deg], Logger Offset, Offset Applied [deg], Date From, Date To anyway and have Offset Applied reported as 0 (if deadband is None) because no offset adjustment is applied on these scenarios, or the applied deadband value if logger offset is None.
- I think that a measurement should never have logger offset as None, but if it is the case on data model the function should print this out. At the moment function is not printing out any info about this (weather e.g. the `apply_wspd_slope_offset_adj` does).

**To Reproduce**
Run code below to show the reported table for each of the scenarios that can happen for a measurement, results of code is shown on screenshot.

```
data = bw.load_csv(bw.demo_datasets.demo_data)
data['Dir78_test'] = data['Dir78mS'].copy()
data['Dir58_test'] = data['Dir58mS'].copy()

test_meas_config_dict = {
    'Dir78mS': [{'name': 'Dir78mS',
   'measurement_type_id': 'wind_direction',
   'height_m': 78.0,
   'logger_measurement_config.slope': 0.07263,
   'logger_measurement_config.offset': 130.87,
   'date_from': '2025-02-20T00:10:00',
   'date_to': None,
   'vane_dead_band_orientation_deg': None}],
  'Dir78_test': [{'name': 'Dir78_test',
   'measurement_type_id': 'wind_direction',
   'height_m': 78.0,
   'logger_measurement_config.slope': 0.07263,
   'logger_measurement_config.offset': None,
   'date_from': '2025-02-20T00:10:00',
   'date_to': None,
   'vane_dead_band_orientation_deg': None}],
 'Dir58mS': [{'name': 'Dir58mS',
   'measurement_type_id': 'wind_direction',
   'height_m': 58.0,
   'logger_measurement_config.slope': 0.07262,
   'logger_measurement_config.offset': 127.75,
   'date_from': '2025-02-20T00:10:00',
   'date_to': None,
   'vane_dead_band_orientation_deg': 127.75}],
  'Dir58_test': [{'name': 'Dir58_test',
   'measurement_type_id': 'wind_direction',
   'height_m': 58.0,
[...]

PR #570, opened 2026-01-08
iss #569 fix table where NaN is present during groupby

olivia-bentley:
Applying fix where current behaviour with groupby where NaN is present is incorrect


BiancaMorandi:
@olivia-bentley can the branch be deleted as well? Did you make these fixes on another branch?

Issue #576, opened 2026-01-13
[_Measurements__get_properties] function called by several measurements functions is not extracting 'column_name' from 'logger_measurement_config'properties from (OB3BM1.5)


BiancaMorandi:
**Describe the bug**
When `measurements.properties` or other linked class function are extracted from an input data model the 'column_name' field from 'logger_measurement_config' table is not extracted and not shown on the properties dictionary. 

**To Reproduce**
You can see this behaviour running following code and the two displayed outputs will look as on screenshots showing that for e.g. 'Dir78mS' the 'column_name' is reported on the `mm1.data_model` but not on the `mm1.measurements.properties` dictionary.

```
import brightwind as bw

mm1 = bw.MeasurementStation(bw.demo_datasets.demo_wra_data_model)

display(mm1.data_model)

display(mm1.measurements.properties)
```

**Expected behavior**
The 'column_name' should be reported on the `mm1.measurements.properties` dictionary as `logger_measurement_config.column_name `field with values as for list of dictionary reported on data model.

**Screenshots**

<img width="809" height="791" alt="Image" src="https://github.com/user-attachments/assets/db7e4010-8418-4cda-806b-c8ff832bddcc" />

<img width="848" height="903" alt="Image" src="https://github.com/user-attachments/assets/a129a17e-40da-48b1-8471-9439c7cbb1ad" />

PR #578, opened 2026-01-19
iss #576 allow return of lists and dictionaries from raised child when requested

olivia-bentley:
To address the issues #576 and . An extra argument was added to `_flatten_dict` and `_raise_child` to give the option whether to remove lists and dictionaries from the child or not. The existing behaviour was to remove all of them


BiancaMorandi:
@stephenholleran this PR is ready for your review/merge

stephenholleran (approved):

PR #582, opened 2026-01-21
[Iss581] Calculate relative humidity from dew point

sararafter:
Added function `calc_rel_humidity_from_dew_point` to calculate relative humidity from dew point temperature and air temperature.

`_calc_water_saturation_vapour_pressure_Pa` function used is part of Issue #549 (so this should be reviewed and merged first).

Currently a value error is raised if dew point temperature is > air temperature as this doesn't make physical sense and results in relative humidity values > 100%. I think ideally if we implement logging it might be better as a warning, as it is not ideal for it to crash the whole function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,9 @@ Additional labels for pre-release and build metadata are available as extensions
 ---
 ## [2.5.0]
 ### New Features and Enhancements
-1. Updated `calc_air_density` to incorporate additional methods. The methods differ in their approach to deriving water vapour pressure using `_calc_water_vapour_pressure_Pa` and include the pre-existing 'IEC' method (default). The two new methods are based on the Herman Wobus approximation (`_calc_water_saturation_vapour_pressure_Pa`) and require either relative humidity or dew point temperature as inputs.([#539](https://github.com/brightwind-dev/brightwind/issues/539)).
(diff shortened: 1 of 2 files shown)


stephenholleran:
> I was previously raising an error for when dew point temp was greater than air temperature as this should be physically impossible (aside from rare cases of supersaturation) and results in relative humidity values of >100%.
> 
> However, I then noticed that (surprisingly) ERA5 and MERRA2 both seem to allow dew point values which are greater than the air temperature at the same timestamp (only very rarely, for example there are 14 timestamps where dew point temp > temp for the entire record on BrightHub for the node nearest the demo mast). Therefore I have removed the error from this PR as we don't want the function to crash every time it is called. After chatting to @BiancaMorandi, I have replaced it with a warning instead.
> 
> @stephenholleran, any thoughts on how to handle measured and/or reanalysis values of relative humidity > 100%?

hmm, how can you have > 100%!! A warning is definitely better than breaking it. 

It seems like you have the warning `msg` in two different places? Probably in different functions. The msg with "_This will result in a calculated relative humidity value greater than 100%._" is good as it tells the user what to expect in the data. You could go further and say to them that they should either remove or set these values to 100%. Is it possible to add this extra info to the first msg?

Why don't we set the values > 100% to = 100%? And in the warning msg tell them that is what we have done because it is impossible to get values over 100%. Is it possible to get values over 100%?

PR #610, opened 2026-05-13
iss609 support nested conditions and time_range_conditions in apply_c…

### LINKED ISSUE #609: [apply_cleaning_rules] BrightHub cleaning rules now allow for multiple conditions and time filtering
**Is your feature request related to a problem? Please describe.**

The BrightHub `/measurement-locations/{uuid}/cleaning-rules` API response has been extended to support richer cleaning rules, but `apply_cleaning_rules()` cannot consume the new structure. The response now contains:

1. **Nested boolean conditions** — `conditions` can wrap children with `and` / `or` / `not` (recursive), instead of being a single flat comparison.
2. **A new `time_range_conditions` block** — sibling of `conditions`, with the same nested structure, but leaves carry an ISO timestamp `value` and a `comparison_operator_id`, evaluated against the timeseries datetime index.
3. **New leaf fields** `measurement_point_uuid` and `statistic_type_id` on `conditions` and `clean_out` items, alongside the existing `assembled_column_name`.

The current implementation in [`apply_cleaning_rules`](brightwind/load/load.py#L2519-L2615) extracts `conditions['assembled_column_name']`, `comparator_value`, and `comparison_operator_id` directly as a flat dict, so any rule using `and` / `or` / `not` or `time_range_conditions` will `KeyError`. The JSON schema at [cleaning_rule.schema.json](brightwind/load/cleaning_rule.schema.json) already permits nested `and` / `or` / `not` in `conditions` but has no entry for `time_range_conditions`, and `rule.additionalProperties: false` will reject any rule that uses it.

Example of a rule the library currently cannot handle:

```python
{
  'rule': {
    'clean_out': [{'assembled_column_name': 'Hum_98m', ...}],
    'conditions': {'comparator_value': 10, 'comparison_operator_id': 1,
                   'assembled_column_name': 'Hum_98m_sd', ...},
    'time_range_conditions': {
      'and': [
        {'comparison_operator_id': 4, 'value': '2016-12-11T12:40', ...},
        {'comparison_operator_id': 2, 'value': '2017-03-20T23:00', ...}
      ]
    }
  }
}
```

**Describe the solution you'd like**

Keep `apply_cleaning_rules()` strictly schema-driven and BrightHub-agnostic. BrightHub-specific normalisation (e.g. UUID → column-name mapping) belongs in `LoadBrightHub`, not in the generic engine.

1. **`apply_cleaning_rules()` — schema-driven evaluator:**
   - Support **full recursive** `and` / `or` / `not` inside `conditions` via a small recursive evaluator that returns a boolean mask aligned to the DataFrame index. The current BrightHub UI only produces single-level `and` of up to 4 leaves, but the library is generic and the recursion is barely more code than handling `and` alone.
   - Support a new top-level `time_range_conditions` block with the same nested structure, evaluated against `df.index`. ANDed with `conditions` and with any `date_from` / `date_to`.
   - Preserve all existing flat-condition behaviour (regression: the demo file and current test must pass unchanged).
   - Accept but ignore `measurement_point_uuid` and `statistic_type_id` — column resolution continues to use `assembled_column_name` only.
   - **Preserve the existing substring-match `clean_out` resolution** at [load.py:2605-2606](brightwind/load/load.py#L2605-L2606) for this release. The substring match is over-permissive (e.g. a `clean_out` of `Spd_40m` could match a sibling sensor like `Spd_40m_S`) and should be replaced with strict expansion to known stat-type suffixes — but that is a breaking change and is tracked separately for the next major release (see below).

2. **Schema updates ([cleaning_rule.schema.json](brightwind/load/cleaning_rule.schema.json)):**
   - Add a `time_condition` definition (separate from the existing `condition`, since the leaf shape differs: `value` is an ISO string, not `assembled_column_name` + `comparator_value`).
   - Add `time_range_conditions` (optional) under `rule.properties` so it passes `additionalProperties: false`.
   - Bump the schema's internal `$version` to `1.1.0`.
   - Do **not** cap `and` at 4 items — the UI's 4-condition limit is a BrightHub product decision, not a library constraint.

3. **Tests ([tests/test_load.py](tests/test_load.py)):**
   - Existing [`test_apply_cleaning_rules`](tests/test_load.py#L56-L112) must pass unchanged as a regression check.
   - New `test_apply_cleaning_rules_nested_and_time_range` using synthetic DataFrames, covering: flat (regression), `and`, `or`, `not`, nested combinations, `time_range_conditions` alone and combined with `date_from`/`date_to`, validation error for malformed rules.

4. **Affected files:** [brightwind/load/load.py](brightwind/load/load.py), [brightwind/load/cleaning_rule.schema.json](brightwind/load/cleaning_rule.schema.json), [tests/test_load.py](tests/test_load.py), [CHANGELOG.md](CHANGELOG.md). Current version `2.7.0_dev` — no bump required.

**Describe alternatives you've considered**

- **Mapping `time_range_conditions` → `date_from` / `date_to` inside `get_cleaning_rules()`.** Tempting because contiguous windows map 1:1, but lossy for `or` / `not` shapes the new structure can express. Rejected — keep the engine schema-driven and let `LoadBrightHub` handle BrightHub-specific concerns separately.
- **Extending the existing `condition` definition with an optional `value` field** instead of adding a separate `time_condition`. Forces an awkward `oneOf` on the leaf and produces less helpful `jsonschema` error messages. Rejected.
- **Implementing only single-level `and`** to match the current UI scope. Rejected — the recursive evaluator is barely more code, and the library is consumed by hand-written JSON and other tooling.
- **Fixing the substring-match `clean_out` resolution in this issue.** Rejected as out of scope — see Follow-ups.

**Follow-ups (separate issues)**

- **Replace substring-match `clean_out` resolution with strict expansion (next major release).** The current logic uses `if column_to_clean in column_name` to expand a single avg-column name to all related stat-type columns. This is over-permissive — a rule listing `Spd_40m` would also sweep in any unrelated column that happens to contain that substring (e.g. `Spd_40m_S`). The fix is to expand each `clean_out` item with `assembled_column_name = name` to `name` itself plus `f"{name}_{stat}"` for each `stat` in the known stat-type suffix list (`sd`, `max`, `min`, `count`, `availability`, `quality`, `sum`, `median`, `mode`, `range`, `gust`, `ti`, `ti30sec`, `text`, `val`) that actually exists in the DataFrame. This is a behavioural change for any existing rule that relied on broad substring matching, so it should land in the next major release with a clear migration note. The existing test at [test_load.py:70-71](tests/test_load.py#L70-L71) and the demo data's CamelCase suffix convention (`Spd80mSStd`, `Spd80mSMax` rather than `Spd80mS_sd`, `Spd80mS_max`) will also need attention as part of that work.

- **`measurement_point_uuid` → `assembled_column_name` resolution in `LoadBrightHub`.** When added, this belongs in a BrightHub-specific normaliser (e.g. `LoadBrightHub.normalise_cleaning_rules(rules, station)`) that runs before passing rules to the generic engine.

**Caveats / open for discussion**

1. **`time_range_conditions` and the UI.** The current UI only produces `and` of column conditions, with no time-range UX. The library will support `time_range_conditions` because the API/schema does, but there is no UI path to produce one today — they would only arrive via hand-written JSON or future UI work. Worth confirming whether there's a planned UI for this so the library and product evolve in step.

2. **Timezone handling in `time_range_conditions` leaves.** The implementation will pass `value` through `pd.Timestamp(value)`. Users supplying naive timestamps against tz-aware indices (or vice versa) will see pandas' standard error. Open: should we normalise to UTC, or trust the caller? Pre-existing `date_from`/`date_to` behaviour is the same, so deferring seems reasonable.

3. **NaN semantics for `operator.ne` (id 6) on float NaN.** Pandas returns `False` for `NaN != x`, which is arguably wrong for "≠". Pre-existing behaviour, out of scope here, but worth a separate issue if it matters in practice.

4. **Schema validation error UX.** `utils.validate_json` prints errors to stdout rather than raising or returning them. Not addressed here, but a clearer validation error path would help users uploading malformed JSON.

**Additional context**

- The existing helper [`_apply_cleaning_rule`](brightwind/load/load.py#L2470-L2516) is only called from `apply_cleaning_rules` itself (confirmed by grep) and will be removed as part of the refactor; its logic is absorbed into the new evaluator.
- Stat-type suffix list (per the BrightHub data model, referenced by the follow-up): `avg`, `sd`, `max`, `min`, `count`, `availability`, `quality`, `sum`, `median`, `mode`, `range`, `gust`, `ti`, `ti30sec`, `text`, `val`. `avg` is the unsuffixed column itself.

dancasey-ie:
## Summary

Updates `apply_cleaning_rules()` to consume the new BrightHub `/measurement-locations/{uuid}/cleaning-rules` API response shape: nested `and`/`or`/`not` boolean conditions and a new optional `time_range_conditions` block. Existing flat-condition rule files continue to work unchanged.

Resolves #609.

## Changes

- **`apply_cleaning_rules()` ([brightwind/load/load.py](brightwind/load/load.py))** — recursive evaluator for nested `conditions` and a new optional `time_range_conditions` block (ISO-timestamp leaves, evaluated against the DataFrame index, ANDed with `conditions` / `date_from` / `date_to`). New `measurement_point_uuid` and `statistic_type_id` fields are accepted but ignored — column resolution still uses `assembled_column_name`. The old private `_apply_cleaning_rule` helper is removed.
- **[brightwind/load/cleaning_rule.schema.json](brightwind/load/cleaning_rule.schema.json)** — new `time_condition` definition + `time_range_conditions` property; `$version` bumped to `1.1.0`.
- **[tests/test_load.py](tests/test_load.py)** — new `test_apply_cleaning_rules_nested_and_time_range` covering flat (regression), `and`, `or`, `not`, nested combinations, `time_range_conditions` alone and combined with `date_from`/`date_to`, validation error, and statistic-type column expansion of `clean_out`.
- **[CHANGELOG.md](CHANGELOG.md)** — entry under `## [2.X.X]`.

## Follow-ups (separate issues)

- Replace substring-match `clean_out` resolution with strict expansion to known stat-type suffixes (breaking — next major release).
- Add a `LoadBrightHub` normaliser that resolves `measurement_point_uuid` → `assembled_column_name` before rules reach the generic engine.

## Test plan
- [ ] `pytest tests/test_load.py::test_apply_cleaning_rules` — regression check passes unchanged.
- [ ] `pytest tests/test_load.py::test_apply_cleaning_rules_nested_and_time_range` — all subtests pass.
- [ ] `pytest tests/test_load.py` — no unrelated regressions.
- [ ] End-to-end: feed a representative BrightHub response (flat + nested `and` + `time_range_conditions`) into `apply_cleaning_rules` against a real DataFrame and verify the mask is correct.

### DIFF
--- a/tests/test_load.py
+++ b/tests/test_load.py
@@ -112,6 +112,185 @@ def test_apply_cleaning_rules():
     assert (data[data_original["T2m"] > 5][date_from:date_to]["Spd80mS"] == "-").all()
 
 
+def _synthetic_cleaning_df():
+    """Deterministic synthetic DataFrame for nested-condition / time-range tests."""
+    index = pd.date_range('2016-01-01', '2016-03-31 23:50', freq='10min')
+    n = len(index)
+    return pd.DataFrame(
+        {
+            'Spd': np.linspace(0, 20, n),
+            'Dir': np.linspace(0, 359, n),
+            'T2m': np.linspace(-10, 30, n),
+        },
+        index=index,
+    )
+
+
+def test_apply_cleaning_rules_nested_and_time_range():
+    df = _synthetic_cleaning_df()
+
+    # 1. Flat condition regression — equivalent to old single-condition behaviour.
+    rules = [{'rule': {
+        'clean_out': [{'assembled_column_name': 'Spd'}],
+        'conditions': {'assembled_column_name': 'Spd', 'comparison_operator_id': 1, 'comparator_value': 10},
+    }}]
+    cleaned = bw.apply_cleaning_rules(df, rules)
+    expected_mask = df['Spd'] < 10
+    assert cleaned.loc[expected_mask, 'Spd'].isna().all()
+    assert cleaned.loc[~expected_mask, 'Spd'].notna().all()
+
+    # 2. AND of two single comparisons — both must be true.
+    rules = [{'rule': {
+        'clean_out': [{'assembled_column_name': 'Spd'}],
+        'conditions': {'and': [
+            {'assembled_column_name': 'Spd', 'comparison_operator_id': 1, 'comparator_value': 10},
+            {'assembled_column_name': 'T2m', 'comparison_operator_id': 1, 'comparator_value': 5},
+        ]},
+    }}]
+    cleaned = bw.apply_cleaning_rules(df, rules)
+    expected_mask = (df['Spd'] < 10) & (df['T2m'] < 5)
+    assert cleaned.loc[expected_mask, 'Spd'].isna().all()
+    assert cleaned.loc[~expected_mask, 'Spd'].notna().all()
+
+    # 3. OR of two single comparisons — either may be true.
+    rules = [{'rule': {
+        'clean_out': [{'assembled_column_name': 'Spd'}],
+        'conditions': {'or': [
+            {'assembled_column_name': 'Spd', 'comparison_operator_id': 1, 'comparator_value': 5},
+            {'assembled_column_name': 'T2m', 'comparison_operator_id': 3, 'comparator_value': 20},
+        ]},
+    }}]
+    cleaned = bw.apply_cleaning_rules(df, rules)
+    expected_mask = (df['Spd'] < 5) | (df['T2m'] > 20)
+    assert cleaned.loc[expected_mask, 'Spd'].isna().all()
+    assert cleaned.loc[~expected_mask, 'Spd'].notna().all()
+
+    # 4. NOT — negation of a single comparison.
+    rules = [{'rule': {
+        'clean_out': [{'assembled_column_name': 'Spd'}],
+        'conditions': {'not': {'assembled_column_name': 'Spd', 'comparison_operator_id': 1, 'comparator_value': 10}},
+    }}]
+    cleaned = bw.apply_cleaning_rules(df, rules)
+    expected_mask = ~(df['Spd'] < 10)
+    assert cleaned.loc[expected_mask, 'Spd'].isna().all()
+    assert cleaned.loc[~expected_mask, 'Spd'].notna().all()
+
+    # 5. Nested AND containing OR.
+    rules = [{'rule': {
+        'clean_out': [{'assembled_column_name': 'Spd'}],
+        'conditions': {'and': [
+            {'assembled_column_name': 'Dir', 'comparison_operator_id': 4, 'comparator_value': 180},
+            {'or': [
+                {'assembled_column_name': 'Spd', 'comparison_operator_id': 1, 'comparator_value': 3},
+                {'assembled_column_name': 'Spd', 'comparison_operator_id': 3, 'comparator_value': 15},
+            ]},
+        ]},
+    }}]
+    cleaned = bw.apply_cleaning_rules(df, rules)
+    expected_mask = (df['Dir'] >= 180) & ((df['Spd'] < 3) | (df['Spd'] > 15))
+    assert cleaned.loc[expected_mask, 'Spd'].isna().all()
+    assert cleaned.loc[~expected_mask, 'Spd'].notna().all()
+
+    # 6. time_range_conditions alone (combined with an always-true conditions block).
+    rules = [{'rule': {
+        'clean_out': [{'assembled_column_name': 'Spd'}],
+        'conditions': {'assembled_column_name': 'Spd', 'comparison_operator_id': 4, 'comparator_value': -1},
+        'time_range_conditions': {'and': [
+            {'value': '2016-02-01T00:00:00', 'comparison_operator_id': 4},
+            {'value': '2016-03-01T00:00:00', 'comparison_operator_id': 1},
+        ]},
+    }}]
+    cleaned = bw.apply_cleaning_rules(df, rules)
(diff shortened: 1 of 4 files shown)


tests/test_load.py
```diff
@@ -112,6 +112,185 @@ def test_apply_cleaning_rules():
...
+
+    # 2. AND of two single comparisons — both must be true.
+    rules = [{'rule': {
+        'clean_out': [{'assembled_column_name': 'Spd'}],
+        'conditions': {'and': [
+            {'assembled_column_name': 'Spd', 'comparison_operator_id': 1, 'comparator_value': 10},
+            {'assembled_column_name': 'T2m', 'comparison_operator_id': 1, 'comparator_value': 5},
+        ]},
+    }}]
+    cleaned = bw.apply_cleaning_rules(df, rules)
```
olivia-bentley, line 152: should we get a clean copy of df for each? Would a fixture be good for this? Just thinking if so many rules are applied what's left that's not NaN at the end is that then not a good test as it may have already removed all that data in a previous test

olivia-bentley: not cleaned inplace so all OK

olivia-bentley (approved):

stephenholleran (approved):

